Issue #48, opened 2020-11-05
Breaking change in quadpy


mancellin:
The optional dependency `quadpy` has had some breaking changes in recent versions. It breaks the quadrature integration feature in version 1.2 of `capytaine` and causes the failure of the tests in  `test_bem_with_quadrature.py`.

Fixing the version of `quadpy` to 0.14 is the best workaround at the moment:
```
pip install quadpy==0.14.11
```

mancellin:
I'm having issues installing that particular version of quadpy in my development setup.
```
ERROR: Could not find a version that satisfies the requirement quadpy<=0.14.11 (from capytaine[develop])
```
If you have the same issue, you can just remove the `quadpy` lines in `setup.py`.
Since #68, the tests are justed skipped if `quadpy` is not installed.
I'm not sure what I'll do with this dependency in the future.

PR #68, opened 2021-04-22
use skipif for optional dependencies

### LINKED ISSUE #67: skip tests dependent on optional packages
In #62, we started using `@pytest.mark.skipif` to skip tests if an optional dependency is not installed. #62 only did this for `pygmsh` and `meshio`. As @mancellin [pointed out](https://github.com/mancellin/capytaine/pull/62#discussion_r618469118), we should also do this for:

- `vtk` (`test_ui_vtk.py` and `test_io_meshes.py`)
- `matplotlib` (`test_matrices.py`)

ryancoe:
fixes #67

### DIFF
--- a/pytest/test_bem_with_quadratures.py
+++ b/pytest/test_bem_with_quadratures.py
@@ -6,14 +6,16 @@ import numpy as np
 import xarray as xr
 import capytaine as cpt
 
-import quadpy
-from packaging import version
-quadpy_version = version.parse(quadpy.__version__)
-if not quadpy_version <= version.parse('0.14'):
+try:
+    import quadpy
+    from packaging import version
+    quadpy_version = version.parse(quadpy.__version__)
+except ImportError:
     quadpy = None
 
-@pytest.mark.skipif(quadpy is None,
-                    reason='quadpy version must be <=0.14, found {:}'.format(quadpy_version))
+
+@pytest.mark.skipif(quadpy is None or not quadpy_version <= version.parse('0.14'),
+                    reason='quadpy not installed or incorrect version (must be <=0.14) ')
 def test_area():
     mesh = cpt.Sphere().mesh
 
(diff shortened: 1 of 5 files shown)


ryancoe:
There are many `asserts` called within a single test function in `test_matrices.py`... I think it would probably be best practice to break these out to be separate tests (can use a `@pytest.fixture` to do the set up). This would also allow us to use the `@pytest.skipif` decorator to skip the `matplotlib` dependent tests. Are you ok with this approach @mancellin (i.e., breaking out all these asserts to separate tests)?

https://github.com/mancellin/capytaine/blob/87d09bf56dcd1d4a242a29805e9d49511b3947e1/pytest/test_matrices.py#L18-L62

mancellin:
> I think it would probably be best practice to break these out to be separate tests

To identify more easily the failing assertion? It makes sense, that's ok with me.
Although, you seem to have two tests with the same name in bf535fa, is it ok with pytest?

ryancoe:
**two tests with same name:** No, I don't think that's OK (good catch!). Fixed with a7520e1808fee7b43d02ec8db87b95b1b848f598

**break out tests:** with bf535fa, I did this just to the minimum level needed to use `@pytest.mark.skipif`. It can be done more (i.e., so that every `assert` statement is its own test. I don't really see to much value in this though.

mancellin:
All right.

The main need for `skipif` might actually be the tests that require quadpy: they could be skipped if the correct version (0.14) is not installed. Cf #48.

pytest/test_bem_with_quadratures.py
```diff
@@ -1,11 +1,19 @@
...
 
+import pytest
 import numpy as np
 import xarray as xr
 import capytaine as cpt
+
 import quadpy
+from packaging import version
+quadpy_version = version.parse(quadpy.__version__)
+if not quadpy_version <= version.parse('0.14'):
```
mancellin, line 12: Maybe we should test both if quadpy is installed and if the version is correct?

ryancoe: Right, so we need an `and` statement

ryancoe: 0e9d111

mancellin:
Thank you!

PR #79, opened 2021-07-26
Bemio -> Capytaine

### LINKED ISSUE #77: Bemio -> Capytaine
I really like the `xarray` `Dataset` format used by `captyaine`. While as reported in #15, we can go from `capytaine` to `bemio`, it would be great to go the other direction too.

michaelcdevin:
/cc @mancellin @ryancoe 

This allows for BEM simulation output files read into Python using the [`bemio`](https://wec-sim.github.io/bemio/) tool to be converted to an `xarray.Dataset` using Capytaine. This allows for more post-processing using Capytaine even if the BEM simulation is handled using other tools.

The bulk of the changes are concentrated in the new [`dataframe_from_bemio`](https://github.com/michaelcdevin/capytaine/blob/f62dfab4ae87e52224b4205ca3946ef656b3a16c/capytaine/io/xarray.py#L364) function and corresponding changes in [`assemble_dataset`](https://github.com/michaelcdevin/capytaine/blob/f62dfab4ae87e52224b4205ca3946ef656b3a16c/capytaine/io/xarray.py#L216) within `xarray.py`.

Currently, this functionality can handle hydrodynamic coefficients and wave diffraction forces. Bemio does not natively import some data that Capytaine calculates, notably wavelength and rigid body mass, so this PR is not comprehensive. However, I'm open to ideas on handling these for a potential PR in the future.

Resolves #77.


capytaine/io/xarray.py
```diff
@@ -238,15 +239,44 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
...
+            if 'bemio.io' in results.__module__:
+                bemio_import = True
+            else:
+                raise TypeError('results must be either of type LinearPotentialFlowResult or a bemio.io object')
+        except:
+            raise TypeError('results must be either of type LinearPotentialFlowResult or a bemio.io object')
+    
+    if bemio_import:
+        records = dataframe_from_bemio(results)
+        all_dofs_in_order = {'Surge': [1., 0., 0., 0., 0., 0.],
```
mancellin, line 262: I don't remember why I used a dict with `None`s as values for `all_dofs_in_order`. A list might have been simpler.
Anyway, do you need the `[1, 0, ...]` values here?

capytaine/io/xarray.py
```diff
@@ -281,40 +311,131 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
 
     # WAVENUMBER
     if wavenumber:
-        dataset.coords['wavenumber'] = wavenumber_data_array(results)
+        if bemio_import:
+            print('Bemio data does not include wavenumber data. wavenumber=True is ignored.')
```
mancellin, line 315: But Bemio stores the frequency, the water depth and the gravity constant, am I right?
So it should be relatively easy to support in the future.

capytaine/io/xarray.py
```diff
@@ -238,15 +239,44 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
...
+            raise TypeError('results must be either of type LinearPotentialFlowResult or a bemio.io object')
+
+    else:
+        try:
+            if 'bemio.io' in results.__module__:
+                bemio_import = True
+            else:
+                raise TypeError('results must be either of type LinearPotentialFlowResult or a bemio.io object')
+        except:
+            raise TypeError('results must be either of type LinearPotentialFlowResult or a bemio.io object')
```
mancellin, line 258: Suggestion: avoid the repetition or the error message in the code by storing it in a variable:
```
error = TypeError("blabla")
if ...:
    raise error
else:
    raise error
```

pytest/test_io.py
```diff
@@ -44,3 +45,10 @@ def test_xarray_dataset_with_more_data():
     ds = cpt.assemble_dataset(results, mesh=True)
     assert 'nb_faces' in ds.coords
     assert set(ds.coords['nb_faces'].values) == set([b.mesh.nb_faces for b in bodies])
+
+
+def test_dataset_from_bemio():
+    with open('bemio_class.pkl', 'rb') as inp:
+        bemio_data = pickle.load(inp)
```
mancellin, line 52: I suppose you used a pickle to avoid explicitly calling the Bemio library. The problem is that it seems to need Bemio to load the pickle.
```
============================== FAILURES ===============================
_______________________ test_dataset_from_bemio _______________________

    def test_dataset_from_bemio():
        with open('bemio_class.pkl', 'rb') as inp:
>           bemio_data = pickle.load(inp)
E           ModuleNotFoundError: No module named 'bemio'

test_io.py:52: ModuleNotFoundError
```
I guess we have no choice but to add Bemio as a (optional?) testing dependency.
If we do so, it might be more robust to load from Bemio data file format rather than from a pickle.

capytaine/io/xarray.py
```diff
@@ -281,40 +311,131 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
...
+                    return next(iter(d.values()))
+                dataset.coords['nb_faces'] = the_only(nb_faces)
+                dataset.coords['quadrature_method'] = the_only(quad_methods)
 
     # HYDROSTATICS
     if hydrostatics:
-        bodies = list({result.body for result in results})
-        dataset = xr.merge([dataset, hydrostatics_dataset(bodies)])
+        if bemio_import:
+            print('Bemio data import being used, hydrostatics=True is ignored.')
```
mancellin, line 350: Suggestion: replace `print` by `LOG.warning`, as used elsewhere in the code.

mancellin (commented):
Thank you for this, it looks good!
The only major issue is with the testing (see comment).

capytaine/io/xarray.py
michaelcdevin: Thanks, this was an oversight on my part. Corrected in commit 0f16541.

capytaine/io/xarray.py
michaelcdevin: Agreed, this makes it much more readable. Fixed in commit 0f16541.

capytaine/io/xarray.py
michaelcdevin: You are correct... I was so caught up in comparing the `capytaine` and `bemio` structures that I glossed over my basic wave theory! Since you effectively already implemented this in `problems_and_results.py`, this functionality was easy to adapt for `dataframe_from_bemio.py` in commit 0f16541. Thanks for catching this!

capytaine/io/xarray.py
michaelcdevin: They are not needed. I initially put these in as placeholders instead of `None`s, as I wasn't sure at the time whether the dict values were used anywhere, and then neglected to change them. This has been corrected in commit 0f16541.

pytest/test_io.py
michaelcdevin: I agree. I suppose if a user is using the Bemio importing functionality, it is reasonable to assume that they already have installed Bemio on their local machine. In lieu of this, 3810be5 changes `test_dataset_from_bemio` to use the [`importorskip`](https://docs.pytest.org/en/latest/reference/reference.html#pytest-importorskip-ref) function to import Bemio if it is installed, and skips the test otherwise. Let me know if you think this is a reasonable solution. Note that the test itself still fails as of this commit due to the values being incorrect; see my comment on the overall PR.

As a potentially relevant side note, the Python distribution of Bemio is no longer supported, and is still written in Python 2, so installing the base repo of Bemio would cause issues with this test. I have created a Python 3 compatible fork of Bemio [here](https://github.com/michaelcdevin/bemio/tree/master-python3), which is what I have been using for testing.

Obviously, since Capytaine requires Python 3.6+, any user using Bemio in conjunction will have already addressed the version incompatibility somehow (using my fork or otherwise), so I don't believe a code solution to this is needed (let me know if you think otherwise). However, once I get around to documentation for this feature, I will be sure to mention this point.

michaelcdevin:
Thanks @mancellin for your feedback, and for your patience with me getting back around to this. I have addressed your comments in my earlier commits, but let me know if the fixes need improvement.

Since I opened this PR, I discovered that I am not correctly dimensionalizing values from Bemio if importing from WAMIT (which uses nondimensionalized hydrodynamic coefficients), and I still need to implement a fix for this. `test_dataset_from_bemio` will fail at the moment as a result of this. I should push the fix in the coming days.

pytest/test_io.py
michaelcdevin: Unit test is now working as of 2574eaf

michaelcdevin:
> Since I opened this PR, I discovered that I am not correctly dimensionalizing values from Bemio if importing from WAMIT (which uses nondimensionalized hydrodynamic coefficients), and I still need to implement a fix for this. test_dataset_from_bemio will fail at the moment as a result of this. I should push the fix in the coming days.

Apologies, the second half of this statement is incorrect; `test_dataset_from_bemio` was failing due to a differing order of omegas, not due to the dimensionalization issue. `test_dataset_from_bemio` is now working as expected as of 2574eaf. The dimensionalization additions still need to be made.

michaelcdevin:
Everything I was wanting to add to this PR has now been added @mancellin. I have not added any documentation for this feature, but I can write it if you would like.

capytaine/io/xarray.py
```diff
@@ -257,15 +257,39 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
     """
     dataset = xr.Dataset()
 
+    error = 'results must be either of type LinearPotentialFlowResult or a bemio.io object'
+    if hasattr(results, '__iter__'):
+        try:
+            if 'capytaine' in results[0].__module__:
+                bemio_import = False
+            else:
+                raise error
```
mancellin, line 266: You forgot the exception, you are raising only the string of the error message.

capytaine/io/xarray.py
```diff
@@ -300,40 +324,184 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
...
     dataset.attrs.update(attrs)
     dataset.attrs['capytaine_version'] = __version__
     return dataset
 
 
+#######################
+#  Import from Bemio  #
+#######################
+
+def dataframe_from_bemio(bemio_obj, wavenumber, wavelength):
```
mancellin, line 387: I suggest we move this function into a separate file `capytaine/io/bemio.py` or something like that.

capytaine/io/xarray.py
```diff
@@ -300,40 +324,184 @@ def assemble_dataset(results: Sequence[LinearPotentialFlowResult],
...
+                temp_dict['influenced_dof'] = dofs
+                
+                if wavenumber:
+                    if temp_dict['water_depth'] == np.infty or omega**2*temp_dict['water_depth']/temp_dict['g'] > 20:
+                        temp_dict['wavenumber'] = omega**2/temp_dict['g']
+                    else:
+                        temp_dict['wavenumber'] = newton(lambda x: x*np.tanh(x) - omega**2*temp_dict['water_depth']/temp_dict['g'], x0=1.0)/temp_dict['water_depth']
+
+                if wavelength:
+                    if temp_dict['wavenumber'] == 0.0:
```
mancellin, line 436: Does this work if `wavenumber=False, wavelength=True`?

mancellin (changes requested):
Thank you, it looks good.
Yes, some documentation is needed, especially if the user needs to figure out the right version of BEMIO to use.
Also, could you add a few lines in the `docs/changelog.rst` file?

capytaine/io/xarray.py
michaelcdevin: Thanks for catching this. Fixed in 8a7d81a.

capytaine/io/xarray.py
michaelcdevin: Done in 231bf6d.

capytaine/io/xarray.py
michaelcdevin: Thanks, this is fixed in 3be53fa.

michaelcdevin:
Okay, I added a section to `docs/outputs.rst` discussing how to use the new feature and which version of Bemio to use. I also updated `docs/changelog.rst` as well.

mancellin:
The file `Bemio_verification_cases/sphere.out` is missing to run the tests.
Otherwise, everything looks good to me :+1:

michaelcdevin:
File added, just forgot to `git add` in the previous commit :)

Thanks for all of your help with this PR @mancellin!

ryancoe:
Hi @mancellin - I just wanted to see if you're indeed planning to incorporate this PR in v1.3? @michaelcdevin and I are working on another project that relies on this change, and I'm trying to decide how to handle the dependencies for that project (i.e., whether to point to this repo or @michaelcdevin's fork).

mancellin:
@ryancoe Yes, this PR should be in v1.3.
I'm a bit busy these days, but I'll release 1.3 at the latest in early October.

mancellin:
@michaelcdevin If we are going to permanently store a relatively large WAMIT result file, it would be nice to get two birds with one stone and also use it to validate the BEM solver. Do you have the parameters used to make this computation?

michaelcdevin:
The .out file I added is identical to the one included in the Bemio repository under [`bemio/tutorials/wamit/sphere`](https://github.com/WEC-Sim/bemio/tree/master/tutorials/wamit/sphere). This directory contains the WAMIT input files used to create the output file, with most of the parameters in the [`sphere.pot`](https://github.com/WEC-Sim/bemio/blob/master/tutorials/wamit/sphere/sphere.pot) and [`sphere.cfg`](https://github.com/WEC-Sim/bemio/blob/master/tutorials/wamit/sphere/sphere.cfg) files.

mancellin:
Alright, thank you, I guess we have all we need to make use of this file. But this is a story for another PR.

PR #80, opened 2021-07-27
fixed FutureWarning in _dataset_from_dataframe function

ckesanapalli:
This resolves the FutureWarning of Pandas.DataFrame.dropna. 

`FutureWarning: In a future version of pandas all arguments of DataFrame.dropna will be keyword-only`.


mancellin:
Thank you!

Issue #81, opened 2021-07-27
Number of free surface mesh panels


marcogambarini:
Hi,
I have found that the free surface elevation evaluation does not seem to work for certain numbers of free surface mesh panels. It looks to me that the problem is in solver.py, specifically in get_potential_on_mesh, where chunk_size is defined and used.

I have found a solution and implemented it locally; unfortunately, I have just recently started using git and my local version of the repo is quite a mess. I may post here the corrected lines if that's ok (I understand that's not good practice, sorry!)

Marco

mancellin:
No problem, if it is just a couple of lines in a single file, I can make the git patch myself.
For small changes, Github also has an option to edit a file online and directly submit a pull request.

marcogambarini:
Ok, I've done that! Thank you.

PR #82, opened 2021-07-27
Potential evaluation for arbitrary number of mesh faces

marcogambarini:
Fixed the indices of faces to extract when mesh.nb_faces > chunk_size. Now the number of mesh faces need not be a multiple of chunk_size.


mancellin:
Thank you!

Closes #81

PR #86, opened 2021-08-09
Improve fill dataset

cmichelenstrofer:
When using _fill_dataset_ users can now:
- include _g_ in the test matrix
- specify the _convention_ and _free-surface_

### DIFF
--- a/capytaine/bem/solver.py
+++ b/capytaine/bem/solver.py
@@ -23,6 +23,7 @@ from datetime import datetime
 from capytaine.green_functions.delhommeau import Delhommeau
 from capytaine.bem.engines import BasicMatrixEngine, HierarchicalToeplitzMatrixEngine
 from capytaine.io.xarray import problems_from_dataset, assemble_dataset, kochin_data_array
+from capytaine.bem.problems_and_results import _default_parameters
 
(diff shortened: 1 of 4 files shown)


mancellin:
Thank you for this!

I agree that `g` should be there, at least for consistency with the other parameters.

Regarding `convention`, I was actually thinking of removing it in the long term. It is hard to maintain and I'm not even sure it is currently correct.

Regarding `free_surface`, your PR has made me aware of a blind spot in Capytaine. More details in #88

cmichelenstrofer:
OK, since `convention` won't be there in the long term I will remove it. Your suggestion in #88 is to only allow `free_surface=0` in DataSets. So I will remove this option too. The only addition from this pull request will be `g` in the test matrix.

mancellin:
Alright, looks good to me.
May I ask you to add a line in the changelog and add a test? (Maybe just adding `g` to one of the existing tests involving `fill_dataset`)

cmichelenstrofer:
I added `g` in a similar way to how `rho` is tested in `test_fill_dataset`. Also added to the changelog.

mancellin:
Thank you!

PR #87, opened 2021-08-09
Fixes #85

### LINKED ISSUE #85: Use pathlib.Path to define filepaths
At least one mesh reader from meshmagick does not accept a file path given as a `Path` from the `pathlib` library.

```python
from pathlib import Path

from capytaine.io.mesh_loaders import load_STL
from capytaine.io.mesh_writers import write_STL
import capytaine as cpt

path = Path()
m = cpt.Sphere().mesh

write_STL(path / "mesh.stl", m.vertices, m.faces)
load_STL(path / "mesh.stl")
```

cmichelenstrofer:
The problem was with VTK. The `setFileName` function cannot handle PosixPath inputs. This affected 5 readers and 2 writers. The solution here is to replace all instances of: 
`SetFileName(filename)`
with
`SetFileName(str(filename))`

### DIFF
--- a/capytaine/io/mesh_loaders.py
+++ b/capytaine/io/mesh_loaders.py
@@ -419,7 +419,7 @@ def load_VTU(filename, name=None):
     vtk = import_optional_dependency("vtk")
 
     reader = vtk.vtkXMLUnstructuredGridReader()
-    reader.SetFileName(filename)
+    reader.SetFileName(str(filename))
     reader.Update()
     vtk_mesh = reader.GetOutput()
 
@@ -451,7 +451,7 @@ def load_VTP(filename, name=None):
     vtk = import_optional_dependency("vtk")
 
     reader = vtk.vtkXMLPolyDataReader()
-    reader.SetFileName(filename)
+    reader.SetFileName(str(filename))
     reader.Update()
     vtk_mesh = reader.GetOutput()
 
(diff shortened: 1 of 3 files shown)


mancellin:
Thank you for handling this!
Would you mind adding a line in the `docs/changelog.rst` file mentioning the bug fix?
Also, since this code comes directly from meshmagick, it would be nice to submit the fix also there.

cmichelenstrofer:
Changelog updated. I believe the problem is with VTK not meshmagick (unless I'm missing something). I took a look at VTK but it is a very large project. I'll mention this as a feature request and hopefully they'll fix it in the future.

cmichelenstrofer:
Posted as Issue in VTK repo: https://gitlab.kitware.com/vtk/vtk/-/issues/18273

mancellin:
Thank you!
Actually, the mesh loaders/writers have been directly borrowed from meshmagick: https://github.com/LHEEA/meshmagick/blob/master/meshmagick/mmio.py This is what I had in mind.
But submitting the issue at the origin in the VTK library is also a good idea.

cmichelenstrofer:
This has been resolved from the VTK side (see issue link above). I would expect it to be available in the next release. At that point we could remove the changes done by this pull request.

mancellin:
Great! Let's keep our patch anyway: it make take some time before the new vtk is widely distributed and it is not really burdening our code.

Issue #27, opened 2020-05-07
Opening hydrostar mesh


martijnwilpshaar:
Hi,

When opening a hydrostar (.hst) mesh, it gives the following error: 

File "C:\Anaconda3\lib\site-packages\capytaine\io\mesh_loaders.py", line 151, in load_HST
    vertices_tmp.append(list(map(float, node.split()[1:])))

AttributeError: 'numpy.ndarray' object has no attribute 'append'

I'm interested in opening this .hst mesh and see the underwater volume (which is necessary for my further calculations). This mesh has been generated by an external company so I cannot share the details unfortunately. 

Could you have a look?

Many thanks!

mancellin:
Yes, there seems to be an issue with the mesh loader.
I'm not familiar with the HST format, so it won't be easy to fix without an example of a failing file.

@frongere Do you remember the logic behind [this section of the code?](https://github.com/LHEEA/meshmagick/blob/37c5b1c0eaa360ebe1ac161022986f46c87b073b/meshmagick/mmio.py#L148)

dirkpir:
I am also having trouble opening a Hydrostar mesh although I get a different error message:

_AssertionError: Faces of a mesh should be provided as positive integers (ids of vertices)_

Replacing the following line of code in **def load_HST**

_return Mesh(vertices, faces-1, name)_

by

_return Mesh(vertices, faces, name)_

fixed this issue when loading the hydrostar mesh (as by taking faces-1 results in a face being represented by a negative integer -1)

However, when trying to solve the diffraction/radiation problem, the GRMS solver did not converge, and the direct linear solver resulted in an empty dataset.

I attached the hydrostar mesh (replacing the .hst extension by .txt to be able to attach it)

[Ship.txt](https://github.com/mancellin/capytaine/files/6998793/Ship.txt)

mancellin:
Indeed, the mesh reader does not seem to be able to read all possible variants of HST files. There are several issues here:

* as you noticed, the mesh reader expects the vertices to be numbered starting from 1, and not from 0. The Hydrostar documentation does not seem to specify how the vertices should be numbered (but the examples there start at 1).
* your panels are defined as triangles of the form
  ```
  panel_index vertex_1 vertex_2 vertex_3
  ```
  but are wrongly interpreted as a list of quadrangles:
  ```
  vertex_1 vertex_2 vertex_3 vertex_4
  ```
  The mesh reader is at fault here.
[...]

PR #90, opened 2021-08-17
New HST mesh reader

mancellin:
Does not solve all issues of #27, but at least returns more helpful error messages.
And it is probably easier to debug and to add feature in comparison with the regular expressions used previously.


mancellin:
TODO: submit to meshmagick?

PR #92, opened 2021-08-20
Updated theory.rst

marcogambarini:
Wrote the sections about forces, dynamic coupling (RAO) and free surface computation.

### DIFF
--- a/docs/theory_manual/theory.rst
+++ b/docs/theory_manual/theory.rst
@@ -71,7 +71,7 @@ The normal velocity on the floating body surface is the input of the problem.
 It depends on the type of problem:
 
 **Radiation problem**:
-    For the radiation problem, the normal velocity on the body surface corresponds to the motion of the body along one of its degree of freedom.
+    For the radiation problem, the normal velocity on the body surface corresponds to the motion of the body along one of its degrees of freedom.
     The resolution of the Laplace problem allows to derive the added mass and the radiation damping associated with this degree of freedom (see also Post-processing_).
 
 **Diffraction problem**:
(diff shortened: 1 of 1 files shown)


mancellin:
Thank you very much!

docs/theory_manual/theory.rst
```diff
@@ -402,15 +402,62 @@ Then other magnitudes such as the Froude-Krylov forces or the added mass can be
...
+
+.. math:: \dfrac{\partial \phi}{\partial t} = - g \eta
+
+which, in frequency domain, is
+
+.. math:: \eta = \dfrac{j \omega}{g} \Phi
+
+For a fully coupled problem (bodies free to oscillate, i.e. diffraction and radiation combined), the free surface elevation can be computed as 
+
+.. math:: \eta = \eta_{\text{incident}} + \eta_{\text{diffracted}} -j \omega \sum_i \eta_{\text{radiated}, i}   X_i
```
mancellin, line 457: Wait. Shouldn't there be also a `-jω` in front of the diffraction term? Isn't the diffraction problem solved with the same conventions?

mancellin: I see what you mean, the `jω` of the diffraction is already in `airy_wave_velocity`.

Actually, it might make sense to add a `jω` directly in https://github.com/mancellin/capytaine/blob/de208c89ddbe5592253128166c32436cfcd8c68a/capytaine/bem/problems_and_results.py#L283 to have the same conventions everywhere. The only problem of doing that is that it would confuse everybody used to the current implementation...

marcogambarini: Yes, that would mean having to correct the computation of added mass and radiation damping accordingly.

PR #96, opened 2021-09-14
Easier developer install

ryancoe:
It appears you can list optional dependencies in `setup.py` under custom headings within the `extras_require` field. This seems like a handy way to list the relatively large number of optional dependencies now in `capytaine`.

This PR implements this optional dependency approach and uses the `develop` flag to invoke these. This will eliminate/reduce the need for folks to hunt these down sequentially and install them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ $(SOFILE): $(F90FILES)
 update_fortran: $(SOFILE)
 
 develop: $(SOFILE)
-	pip install -e .
+	pip install -e .[develop]
 
 test: develop
 	python -m pytest
--- a/docs/developer_manual/installation.rst
+++ b/docs/developer_manual/installation.rst
@@ -11,7 +11,7 @@ It is recommended to use a `conda environment`_.
 
 Ensure that Numpy is installed in your enviroment::
 
-    conda install numpy
+    conda install numpy pip
 
 You'll also need a Fortran compiler:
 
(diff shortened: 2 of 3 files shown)


ryancoe:
Not sure why this CI build failed (Python 3.7 fails, but 3.9 succeeds)... ?

mancellin:
> Not sure why this CI build failed (Python 3.7 fails, but 3.9 succeeds)... ?

Looks like a random internet connection error. Probably nothing to see here.

mancellin:
Thank you, it is indeed more convenient.

I'm a bit reluctant to keep only the `make` invocation in the documentation, in case somebody wants to build the code without `make` (e.g. on Windows). But it is such a niche case that it might not be worth it...

ryancoe:
I agree... I was revising the documentation in that section and originally kept everything, but that felt too complicated. One idea might be to simply point people to open up the `Makefile` if they want to understand the separate steps.

mancellin:
Maybe, we can just provide developers installation instructions for Linux, MacOS and Windows Subsystem for Linux. I'm not even sure that the current Windows instructions are working...

Besides, the Makefile might be too complicated: I don't believe `F90FILES` and `SOFILES` are actually used. I'll check that.

mancellin:
@ryancoe Could you take a quick look at the latest commit here? If it looks good to you, I'll merge this PR.

ryancoe:
Looks good to me. One thought I had was whether we should have `sphinx` and `sphinxcontrib-proof` listed in `develop` packages to allow people to easily compile the docs. I guess I can see this both ways... what do you think @mancellin?

mancellin:
Yes, it would make sense to add `sphinx` and `sphinxcontrib-proof`.

Maybe we could have two different `extras_require`ments: one with the optional dependencies useful for users (`vtk`, `meshio`, ...) and one with the development tools (`pytest`, `sphinx`, ...).

ryancoe (commented):
This will allow an "advanced" user to call `pip install -e .[extra]`, and have the full set of packages. It may also be helpful for CI/CD (not quite sure how this is being done at the moment).

mancellin:
Thank you!

> It may also be helpful for CI/CD (not quite sure how this is being done at the moment).

This is done with the `conda` and `conda-forge` toolchain, which do not use the `setup.py`, but
https://github.com/mancellin/capytaine/blob/6290cc0378ec769fdaa3220cd11a772793bab93d/conda/meta.yaml#L29-L42

I'll see what I can do to improve that.

Issue #32, opened 2020-05-21
funny behavior in v1.2


ryancoe:
I'm working on an analysis of the an array of five wave star devices, and I think I've found some odd behavior in v1.2.

![image](https://user-images.githubusercontent.com/5589887/82608497-0de95e00-9b78-11ea-9396-a26d9b7392c6.png)

```python
from capytaine import FloatingBody
import logging
import numpy as np
import capytaine as cpt
import meshmagick.mesh as mm
import meshmagick.hydrostatics as hs
import matplotlib.pyplot as plt
import scipy.io as sio
import sys
import os
import meshio as mio
import pygmsh
import xarray as xr
from capytaine.io.mesh_writers import write_mesh 



def waveStarArray(alpha, f_range=None, fbname='waveStar', runBem=True, 
                  showGui=False, createAnim=False):
    
    if f_range is None:
        f_range = np.linspace(0.2, 4, 25)
        
    w_range = 2* np.pi * f_range    

    # device dimensions 
    D = 0.254
    rho = 1000
    dy1 = 0.75*alpha
    dy2 = 0.7*alpha
    dx = 1.6*alpha**2
    
    # define single device
    geom = pygmsh.opencascade.Geometry()
    sphere = geom.add_ball([0,0,0], D/2)
    
    # mesh and write mesh files
    mshRefFactor = 1
    mshArgs = ['-clscale', str(mshRefFactor),
               '-clcurv', str(360/50)]
    mesh = pygmsh.generate_mesh(geom,
                                dim=2,
                                extra_gmsh_arguments=mshArgs,
                                remove_lower_dim_cells=True,
                                geo_filename=fbname + '.geo',
                                msh_filename=fbname + '.stl',
                                mesh_file_type="stl")
    mesh.write(fbname + '.vtk')
    
    # import mesh into Capytaine
    fb = cpt.FloatingBody.from_file(filename=fbname+'.stl', name=fbname)
    fb.keep_immersed_part()
    fb.add_translation_dof(direction=(0,0,1))
    
    # my_axis = cpt.meshes.geometry.Axis((0, 1, 0), 
                                        # point=(-0.437537427, 0, 0.281))
    my_axis = cpt.meshes.geometry.Axis((0, 1, 0), 
                                        point=(0, 0, 0))
    
    # import into meshmagick
    hsb = hs.Hydrostatics(mm.Mesh(fb.mesh.vertices, fb.mesh.faces), 
                          rho_water=rho)
    
[...]

PR #58, opened 2021-04-13
update numpy dtypes per guidance from numpy/pytest

ryancoe:
Updated dtypes per https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. This is similar to https://github.com/LHEEA/meshmagick/pull/23.

All tests pass

```
$ pytest
========================================================================= test session starts ==========================================================================
platform darwin -- Python 3.9.2, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
rootdir: /Users/rcoe/capytaine
plugins: hypothesis-6.9.1
collected 100 items                                                                                                                                                    

pytest/test_bem_engines.py ..                                                                                                                                    [  2%]
pytest/test_bem_green_functions.py ....                                                                                                                          [  6%]
pytest/test_bem_hierarchical_toeplitz_matrices.py ..........                                                                                                     [ 16%]
pytest/test_bem_linear_combination_of_dofs.py ..                                                                                                                 [ 18%]
pytest/test_bem_problems_and_results.py .........                                                                                                                [ 27%]
pytest/test_bem_solver.py ...                                                                                                                                    [ 30%]
pytest/test_bem_with_quadratures.py ..                                                                                                                           [ 32%]
pytest/test_bodies.py ..........                                                                                                                                 [ 42%]
pytest/test_bodies_predefined.py .....                                                                                                                           [ 47%]
pytest/test_consistency_with_Nemoh_2.py .......                                                                                                                  [ 54%]
pytest/test_io.py .                                                                                                                                              [ 55%]
pytest/test_io_meshes.py .                                                                                                                                       [ 56%]
pytest/test_matrices.py ...............                                                                                                                          [ 71%]
pytest/test_meshes.py ................                                                                                                                           [ 87%]
pytest/test_meshes_collections_and_symmetries.py ......                                                                                                          [ 93%]
pytest/test_meshes_geometry.py .....                                                                                                                             [ 98%]
pytest/test_post_pro_kochin.py .                                                                                                                                 [ 99%]
pytest/test_ui_vtk.py .                                                                                                                                          [100%]

========================================================================= 100 passed in 8.91s ==========================================================================
```

PR #97, opened 2021-09-15
Generalize GDF read to accept multiple formats

michaelcdevin:
This mirrors my open PR for Meshmagick [#32](https://github.com/LHEEA/meshmagick/pull/32). Since this code for Meshmagick and Capytaine is basically identical, it makes sense to apply this fix to Capytaine as well.

Previously, load_GDF only accepted GDF files that had each vertex on its own line, so the coordinates for each face were 4x3. However, GDF files can also have other formats, such as all four vertices sharing a line, so each face is 1x12. This makes it so load_GDF can parse any format of GDF file regardless of line breaks, as long as each face is separated by a line break.


capytaine/io/mesh_loaders.py
```diff
@@ -671,16 +671,23 @@ def load_GDF(filename, name=None):
     line = ifile.readline().split()
     nf = int(line[0])
 
-    vertices = np.zeros((4 * nf, 3), dtype=float)
-    faces = np.zeros((nf, 4), dtype=int)
+    vertices = np.zeros((4 * nf, 3), dtype=np.float)
+    faces = np.zeros((nf, 4), dtype=np.int)
```
mancellin, line 675: Changing back #58?

mancellin:
Looks good to me. Would you mind adding a line in the changelog?

capytaine/io/mesh_loaders.py
michaelcdevin: My bad, switched back.

PR #98, opened 2021-09-17
fix example bug: plot_influence_matrix

ryancoe:
I don't fully understand why, but it looks like this stopped working due to d7de51da2a60d0b315258deb2152d9c561c6febb

### DIFF
--- a/docs/user_manual/examples/plot_influence_matrix.py
+++ b/docs/user_manual/examples/plot_influence_matrix.py
@@ -22,7 +22,7 @@ S, K = engine.build_matrices(
 
 # Plot the absolute value of the matrix S
 import matplotlib.pyplot as plt
-plt.imshow(abs(S))
+plt.imshow(abs(S.full_matrix()))
 plt.colorbar()
 plt.title("$|S|$")
 plt.tight_layout()


ryancoe:
It seems that the object `S` was changed in d7de51d; we must now convert it from a `BlockSymmetricToeplitzMatrix` to an `np.ndarray` before plotting. Again, I don't fully understand how this bug got introduced, so may be good for @mancellin to review and see if there are likely other consequences.

https://github.com/mancellin/capytaine/blob/e140bf7c5894dcb7ac5653dbed5e60015ce0c0a2/docs/user_manual/examples/plot_influence_matrix.py#L16-L29

mancellin:
Good catch, thank you.
The default behavior has changed in d7de51d to use the reflection symmetry (xOz plane) of the horizontal cylinder. This is better in most cases because only half of the coefficients are required, but it may break some advanced use cases such as this example...
The fix is either to "fill" the sparse matrix (i.e. copy the computed half of the coefficient to get the other half) as you did
```
S = S.full_matrix()
```
or to disable the symmetry of the cylinder
```
cylinder = cpt.HorizontalCylinder(..., reflection_symmetry=False)
```

mancellin:
Fixes error introduced in #91

Issue #103, opened 2021-09-29
Meshmagick 3.0 and stiffness matrix


francesco1788:
Good day,
In the example "boat_animation.py" and "hydrostatics.py", the hydrostatics have been calculated using the attribute "Hydrostatics" of the module meshmagick.hydrostatics.
I just wanted to bring to your attention that in the latest (3.0) version of meshmagick, this attribute does not exist anymore; so the example codes will only run with a meshmagick version equal to 2.1 or older.
I thought that they had replaced it with the attribute 'compute_hydrostatics', but it still does not run properly.
Thanks

mancellin:
Thank you for reporting this!

It seems that the old hydrostatics module is still available under the name `meshmagick.hydrostatics_old`, so it is still possible to run the current examples with Meshmagick 3.0.
If you take the time to learn the new module, please come back to help us update the examples :)

ryancoe:
Before closing #106, we can update `boat_animation.py` to use the new methods internal to `capytaine` instead of those from `meshmagick`.

PR #106, opened 2021-09-29
added the Interia mass matrix, hydrostatic stiffness and other hydrostatic functions

ckesanapalli:
I added functions to compute volumes using x, y, z components, mass, waterplane area, the center of buoyancy, Interia mass matrix (experimental), and hydrostatic stiffness matrix for all type of DOF including flexible and compressible mode (requires divergence of the modes are input for compressible mode case). 

Regarding the inertia mass matrix, I computed using the gauss divergence theorem and compared it with meshmagick, apart from Ixy, Iyz, Ixz rest of the terms are in good agreement.


ryancoe:
@akeeste - thought you might be interested in this

mancellin:
Thank you very much @ckesanapalli for working on this!

It looks good, but I need to ask you is to add some validation tests (e.g. in a new file `pytest/test_hydrostatics.py`).
Ideally, it should include at least:
* a test on a single 6-dof rigid body,
* a test on a combination of bodies such as:
https://github.com/mancellin/capytaine/blob/6290cc0378ec769fdaa3220cd11a772793bab93d/docs/user_manual/examples/multibody.py#L29
* a test on a flexible body.

> Also changed the sign of the radiation damping term in impedance function

If you don't mind, I'd prefer to have separate issues in different Pull Requests.

ckesanapalli:
Sorry for the late update. I have added some more functions similar to `meshmagick.hydrostatics.compute_hydrostatics` function. After comparing with meshmagick, some of the hydrostatic stiffness matrix 
 terms i.e., C44, C55 is not always matching with the meshmagick results for all cases. 

I think this is due to use of rigid DOFs to compute hydrostatic coefficient  instead of directly using integrals  ([WAMIT guide](https://www.wamit.com/manualupdate/history/V70_manual_old.pdf) pg 24). 

The equation I used to compute hydrostatic coefficients is referred from 
[`Newman, John Nicholas. "Wave effects on deformable bodies." Applied ocean research 16.1 (1994): 47-59.`](http://resolver.tudelft.nl/uuid:0adff84c-43c7-43aa-8cd8-d4c44240bed8) Eq-2.17.

mancellin:
> After comparing with meshmagick, some of the hydrostatic stiffness matrix terms i.e., C44, C55 is not always matching with the meshmagick results for all cases.
> 
> I think this is due to the use of rigid DOFs to compute hydrostatic coefficient instead of directly using integrals ([WAMIT guide](https://www.wamit.com/manualupdate/history/V70_manual_old.pdf) pg 24).

No problem for me if two different expressions give results with a 1% difference on a given mesh.
Could you check maybe if this error become smaller when the mesh is finer?

mancellin:
Looks good to me overall!

Can I also ask you to add some documentation for the new methods?
At the very least add an item in `docs/changelog.rst`. Maybe add a few lines in `docs/user_manual/mesh.rst` or add a new `docs/user_manual/hydrostatics.rst`?

ckesanapalli:
I have added in `docs/user_manual/tutorial.rst` and  `docs/user_manual/hydrostatics.rst` page in the documentation.

ckesanapalli:
Newman (1994) formula is not 'complete' as recovering the rigid body terms is not possible. https://doi.org/10.1115/1.3058702.

An alternative is to use the general equation of hydrostatic and  restoring coefficient for rigid modes and use Neuman method for elastic modes. I added that in my latest commit.

I also added an example in `docs/user_manual/hydrostatics.rst` comparing some hydrostatic parameters of Capytaine, Meshmagick with analytical results.

ckesanapalli:
Also I have added the results of comparison of Capytaine, Meshmagick and Analytical cases to `docs/user_manual/hydrostatics.rst`.

mancellin:
Nice work! Thank you!

> Newman (1994) formula cannot properly represent the hydrodynamic stiffness for the rigid body since it doesn't have the body weight component. There are other methods but they require gradient of DOFs. https://doi.org/10.1115/1.3058702.

So, if I understand correctly, the generic formulation is missing a term of the expression for the sake of simplicity. For rigid body dofs, it is exactly computed, hence the small discrepancy between the two.

Your criterion for detecting rigid body dofs could maybe be improved: in case of multi-rigid-body problems, the dofs might be named `body_1__Heave`, `body_2__Surge`, ...
By the way, this could be worth adding to the tests: if you have
```
sphere.add_all_rigid_body_dofs()
cylinder.add_all_rigid_body_dofs()
both = sphere + cylinder  # 12-dof body
```
I guess the stiffness matrix should be a 12x12 matrix with the 6x6 stiffness matrices of each body on the diagonal.

mancellin:
There are a few minor user-interface issues I still need to think about.

For instance, the convention in most of Capytaine is to access values with `object.property` or `object.property()` rather than `object.get_property()`, so I'm wondering if we should do the same here.

Also, there is an (undocumented?) feature that if a body has a `center_of_mass` attribute, the rigid body rotation dofs are defined around this point. I'm not sure how to integrate that with the new hydrostatics methods.

ckesanapalli:
> For instance, the convention in most of Capytaine is to access values with `object.property` or `object.property()` rather than `object.get_property()`, so I'm wondering if we should do the same here.

That is a great idea. I will change the functions to properties. 

> Also, there is an (undocumented?) feature that if a body has a `center_of_mass` attribute, the rigid body rotation dofs are defined around this point. I'm not sure how to integrate that with the new hydrostatics methods.

I didn't know that there is a COG property in capytaine. I should have searched for it. I will change the function to include the COG property by default but also give an option to assign COG. 


> Besides, should `get_hydrostatic_stiffness()` and `get_rigid_dof_mass()` return a bare `numpy` array or a labelled `xarray` DataArray?

I can return xarray for the `get_hydrostatic_stiffness()` since it is directly dependent on the `body.dofs`. 

But the `get_rigid_dof_mass()` always returns 6 x 6 matrix and independent of the `body.dofs`. I don't think it will be consistent to use xarray for `get_rigid_dof_mass()` when user inputs different `body.dofs`

ckesanapalli:
https://github.com/mancellin/capytaine/pull/106#issuecomment-1076477516

Regarding the 12 x 12 stiffness matrix, the current `get_hydrostatic_stiffness()` function can return 12 x 12 stiffness matrix as long as the inputted DOFs are correctly assigned. 

But I will add an example and test illustrating that.

ryancoe:
Another minor note: `boat_animation.py` should probably be updated to use this internal module instead of `meshmagick`, right (see https://github.com/ckesanapalli/capytaine/pull/1)?

https://github.com/mancellin/capytaine/blob/d293b12366edf75d98ff81506d7e1d346a4aeeb7/docs/user_manual/examples/boat_animation.py#L24-L64

ckesanapalli:
Good idea. I will compare new functions results with the meshmagick and analytical case and edit the `boat_animation.py `  to include new functions.

ckesanapalli:
Currently I did a comparison with the stiffness matrices of each body with the merged body. 

```python
import json
from pathlib import Path

import capytaine as cpt
import numpy as np
from hydrostatics import *

# def test_two_body_rigid_dofs():
    
vert_cyl = cpt.VerticalCylinder(
                length=10.0, radius=5.0,
                center=(10.0,0.0,0.5), 
                nr=10, nx=10, ntheta=10,
            ).keep_immersed_part()
    
sphere = cpt.Sphere(
            radius=100,
            center=(0,10,1),
            nphi=50, ntheta=50,
        ).keep_immersed_part()

hori_cyl = cpt.HorizontalCylinder(
                length=5.0, radius=1.0,
                center=(0,10,-2),
                nr=20, nx=20, ntheta=10, 
            ).keep_immersed_part()

vert_cyl = new_float(mesh=vert_cyl.mesh, name="my buoy")
sphere = new_float(mesh=sphere.mesh, name="my buoy")
hori_cyl = new_float(mesh=hori_cyl.mesh, name="my buoy")

vert_cyl.center_of_mass = vert_cyl.center_of_buoyancy
sphere.center_of_mass = sphere.center_of_buoyancy
hori_cyl.center_of_mass = hori_cyl.center_of_buoyancy

vert_cyl.add_all_rigid_body_dofs()
sphere.add_all_rigid_body_dofs()
hori_cyl.add_all_rigid_body_dofs()

all_bodies = vert_cyl + sphere + hori_cyl

all_bodies = new_float(mesh=all_bodies.mesh, name="my buoy")

all_bodies.center_of_mass = all_bodies.center_of_buoyancy

# Adding the rigid DOFs of the vertical cylinder, sphere, and horizontal cylinder
# to `all_bodies`

def add_rigid_dofs(all_bodies, sub_body, prefix):
    zero_dof = np.zeros_like(all_bodies.mesh.faces_centers)
    all_bodies.dofs.update({f"{prefix}_{dof_name}": zero_dof.copy() for dof_name in sub_body.dofs})
        
    for sub_face_idx, sub_face_center in enumerate(sub_body.mesh.faces_centers):
        all_face_idx = np.where((sub_face_center==all_bodies.mesh.faces_centers).all(axis=1))[0][0]
        for dof_name, dof in sub_body.dofs.items():
            all_bodies.dofs[f"{prefix}_{dof_name}"][all_face_idx] = dof[sub_face_idx]
            
            # print(f"{prefix}_{dof_name}", all_face_idx, all_bodies.dofs[f"{prefix}_{dof_name}"][all_face_idx])

all_bodies.dofs={}
add_rigid_dofs(all_bodies, vert_cyl, "vert_cyl")
add_rigid_dofs(all_bodies, sphere, "sphere")
add_rigid_dofs(all_bodies, hori_cyl, "hori_cyl")


# Computing Hydrostatic Stiffness
vert_cyl_hs = vert_cyl.hydrostatic_stiffness_xr()
sphere_hs = sphere.hydrostatic_stiffness_xr()
hori_cyl_hs = hori_cyl.hydrostatic_stiffness_xr()
all_bodies_hs = all_bodies.hydrostatic_stiffness_xr()


# Comparing the Results
def compare_hs(all_bodies, sub_body_hs, prefix):
    print(f"(Influenced DOF, Radiating DOF) - {prefix}, All bodies")
    for influenced_dof_name in sub_body_hs.influenced_dof.values:
        for radiating_dof_name in sub_body_hs.radiating_dof.values:
            vert_cyl_hs_value = sub_body_hs.sel(influenced_dof=influenced_dof_name, 
                                            radiating_dof=radiating_dof_name).values
            all_bodies_hs_value = all_bodies_hs.sel(influenced_dof=f"{prefix}_{influenced_dof_name}", 
                                            radiating_dof=f"{prefix}_{radiating_dof_name}").values
            if not np.isclose(vert_cyl_hs_value, all_bodies_hs_value, rtol=1e-3, atol=1e-4):
                print(f"({influenced_dof_name}, {radiating_dof_name}): {vert_cyl_hs_value:.4e}, {all_bodies_hs_value:.4e}")
    print("="*30)

compare_hs(all_bodies, vert_cyl_hs, "vert_cyl")
compare_hs(all_bodies, sphere_hs, "sphere")
compare_hs(all_bodies, hori_cyl_hs, "hori_cyl")
```

The comparison of the each body stiffness matrix with the merged body stiffness matrix 
NOTE: I only printed the values are the not matching.
```python
(Influenced DOF, Radiating DOF) - vert_cyl, All bodies
(Heave, Pitch): -7.2053e+06, -9.9295e-10
(Pitch, Heave): -7.2053e+06, -1.0104e-09
(Pitch, Pitch): 7.6119e+07, 4.0665e+06
(Pitch, Surge): 0.0000e+00, 3.2424e+06
(Roll, Sway): 0.0000e+00, -3.2424e+06
==============================
(Influenced DOF, Radiating DOF) - sphere, All bodies
(Heave, Roll): 3.0708e+09, -1.7125e-06
(Pitch, Surge): 0.0000e+00, 2.0158e+10
(Pitch, Yaw): -7.1614e-05, 7.5269e-05
(Roll, Heave): 3.0708e+09, -1.7125e-06
(Roll, Roll): 7.9465e+11, 7.6344e+11
(Roll, Sway): 0.0000e+00, -2.0158e+10
==============================
(Influenced DOF, Radiating DOF) - hori_cyl, All bodies
(Pitch, Surge): 0.0000e+00, 1.4411e+05
(Roll, Sway): 0.0000e+00, -1.4411e+05
==============================
```

ckesanapalli:
Changed the function names, mass and stiffness function outputs xarray dataarray. Updated the documents including the `boat_animation.py` file.

mancellin:
> ```python
> # Adding the rigid DOFs of the vertical cylinder, sphere, and horizontal cylinder
> # to `all_bodies`
> def add_rigid_dofs(all_bodies, sub_body, prefix):
>   [...]
> ```

Something of this kind is already happening automatically when you `+` the bodies together, isn't it?

> Currrent all_bodies.hydrostatic_stiffness is not able to detect multi-body rigid DOFs i.e,. body_1_Surge, body_1_Heave, body_2_Sway

I guess you could detect if a dof name ends with a rigid-body-dof name.
But then you would need to figure out exactly what is `body_1` to compute its rigid body properties.
Maybe we are just hitting the limits of Capytaine data model for multi-bodies, and it would not be worth it.

> (Heave, Roll): 3.0708e+09, -4.2812e-07

It seems like a huge difference. But I guess that both numbers are small with respect to other coefficients in the matrix?

> But the get_rigid_dof_mass() always returns 6 x 6 matrix and independent of the body.dofs. I don't think it will be consistent to use xarray for get_rigid_dof_mass() when user inputs different body.dofs

An idea (I'm not so sure about it): we always return a ndof × ndof labelled array.
For the dofs that are detected as being rigid body dofs, the matrix is filled with the computed coefficient.
When other dofs are detected, a warning is displayed and the coefficients in the matrix are NaN.

ckesanapalli:
> For the dofs that are detected as being rigid body dofs, the matrix is filled with the computed coefficient.
> When other dofs are detected, a warning is displayed and the coefficients in the matrix are NaN.

Done.

mancellin:
Thanks!

One more nitpick to polish the PR:
I don't like the fact that a single concept can have several names.
For instance, the `center_of_mass` is named `cog` in the `compute_hydrostatics()` dictionnary. Or the `hydrostatic_stiffness` is named `stiffness_matrix` in this dict.
To make it easier to users, I think we should uniformize that.

ckesanapalli:
> To make it easier to users, I think we should uniformize that.

I totally agree. 

I previously choose the variable names in the `compute_hydrostatics()` same as the meshmagick output for comparison and similarity. 

For now, I will change `cog->center_of_mass` and `stiffness_matrix->hydrostatic_stiffness` in `compute_hydrostatics()` dict. 

If you have any suggestions for others variables. I change there names too.

mancellin:
> If you have any suggestions for others variables. I can change those names too

I can see a few others in `compute_hydrostatics()`. I can also change them myself.

Although, I'm not sure which name to keep for the metacentric parameters. Are the short names ('bmt', 'bml', 'gmt', 'gml') explicit enough for most users to immediately understand what they are? (I'm not from a marine engineering background.)

mancellin:
Fix #103

ckesanapalli:
BM_T, BM_L and and GM_T, GM_L are the convention in naval architecture. But for explicitness, I changed the property names.

mancellin:
Thank you @ckesanapalli 
I'm doing some polishing today (adding a few tests, fixing edge case bugs and improving integration with the rest of the package) and I expect to merge the PR in the upcoming days.

mancellin:
@ckesanapalli : I'm merging PRs by squashing all commits into one. The drawback is that the master branch on your fork might now be diverging from mine. I would recommend to create a dedicated branch for a each PR to avoid this in the future.

PR #109, opened 2021-10-29
Fix various typos

luzpaz:
Found via `codespell -q 3 -L ba,informations,inout,ist,ment,noe,periode,sur`

### DIFF
--- a/capytaine/bodies/bodies.py
+++ b/capytaine/bodies/bodies.py
@@ -76,7 +76,7 @@ class FloatingBody(Abstract3DObject):
 
         import meshio
         if not isinstance(mesh, meshio._mesh.Mesh):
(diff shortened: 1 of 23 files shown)


luzpaz:
bump

luzpaz:
@mancellin Do you think you'd consider fixing the source typo for
`lamda`->`lambda` ?

mancellin:
Yes, sorry for the delay. The breakage of the CI is clearly unrelated to your PR.

> Do you think you'd consider fixing the source typo for lamda->lambda ?

Given that `lambda` is a Python keyword so we can't use it directly and `lamda` is a rare but valid alternative spelling, I believe it is fine the way it is.
If I were to change it, I would use `λ` :)

Issue #121, opened 2022-01-10
Color mesh faces based on value


marcogambarini:
Hi,
I had to write a small script to color mesh faces with values from a vector. Could it be interesting/useful to add it as an extension to function `show_matplotlib` (e.g. to plot pressure distributions)?

marcogambarini:
Ok, I will work on it in the next days!

PR #122, opened 2022-01-16
Plotting scalar field over mesh

marcogambarini:
I modified function show_matplotlib so that a scalar field can be plot over the mesh (see issue #121).

Usage example:
[test-showmpl.py.txt](https://github.com/mancellin/capytaine/files/7877826/test-showmpl.py.txt)
Should I add this to the tests?

Let me know if something needs to be changed.

Good evening!


mancellin:
Thank you. It looks good, I only have a couple of minor issues:

1) I'm not fond of the `fig` argument. It may cause a lot of unpredictable behavior if the user passes only `ax`, or only `fig`, or if they are not consistent with each other. I suggest we keep only `ax` for two use cases:
- Either the user does not provide the axes and we create them for her, including the colorbar.
- Or the user wants some custom layout, she gives an `ax` and we let her handle the colorbar herself.

2) Maybe the name `field` could be made less ambiguous. Something like `color_field` or `scalar_field_on_mesh` or something like that?

> Should I add this to the tests?

Yes. The correctness of the result of display functions is hard to automatically test, but we could at least check that it runs without error.
Could you please also add a line at the top of `docs/changelog.rst`?

marcogambarini:
Ok, I have fixed the function to avoid the `fig` argument and added the line to `docs/changelog.rst`. 
I am not sure about the structure of the test, let me know if it needs to be modified.

mancellin:
I did some minor editing of the test and will now merge into master. Thank you for this PR!

mancellin:
@marcogambarini Beware that the master branch on your fork is probably now out-of-sync with the master branch of the main repository. You might need to force push on it to fix that. In the future, I'd recommend doing the PR from a new dedicated branch that can be deleted or forgotten once the PR is merged.

marcogambarini:
Ok, thank you for the suggestion and for your patience!

PR #123, opened 2022-01-21
function to compute IRF kernels

ryancoe:
- [x] radiation
- [ ] excitation
- [ ] update docs

### DIFF
--- a/capytaine/post_pro/irf.py
+++ b/capytaine/post_pro/irf.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+# coding: utf-8
+"""Experimental function to compute the impendance."""
+# Copyright (C) 2017-2019 Matthieu Ancellin
+# See LICENSE file at <https://github.com/mancellin/capytaine>
+
+import logging
+
+import numpy as np
+import xarray as xr
+
+LOG = logging.getLogger(__name__)
+
+
+def rad_kernel(dataset, time_series):
+	"""Impulse response function kernel (see, e.g., Falnes)
+    
+    @book{falnes2002ocean,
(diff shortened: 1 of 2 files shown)


mancellin:
Thank you Ryan!

I have to set a limit to what should be included in Capytaine and what should not, because the package cannot grow indefinitely. I think it would make sense to draw the line between frequency domain and time domain, and keep only frequency domain stuff.
But at the same time, Capytaine is meant to be a successor of Nemoh and IRF were in Nemoh, so I can make an exception for this one.

mancellin:
It is basically a Fourier transform, isn't it? Would something like `scipy.fft` also do the job?

ryancoe:
Scope creep: Agreed. I wasn't going to submit it here, but then I saw that the functionality was indeed in NEMOH. You're call if you want to keep it out.

Fourier: yes, you're quite right. I will probably rewrite it to use the `ifft`

mancellin:
Hi @ryancoe

What are your plans regarding this?

I believe now that it would be useful to have some basic frequency to time domain conversion tools, at least for teaching purposes. Maybe this could be the IRF.

salhus:
Thanks Ryan! 
Hi @ryancoe @mancellin, 
I think I can take a stab at it. Looks like the radiation IRF is complete, excitation IRF needs to be developed? 
Or you want to rehash them both to use IFFT?

Cheers,
sal

mancellin:
>  Or you want to rehash them both to use IFFT?

That is up to you. Using the Fast Fourier Transform might be faster, but I don't think performance is really an issue there. So use the method that is the most usable/readable.

mancellin:
Closing this since I don't think anybody really has time for that anytime soon. Feel free to reopen or to open another PR.

Issue #100, opened 2021-09-23
Haskind relation


marcogambarini:
Hi! Could it be interesting/useful to use the Haskind relation for simplifying some of the computations? (or maybe to add it as an example)
I have written a small script that implements it, I have attached it. The only thing that puzzles me right now is the sign of forces (comparing, for example, my computation with line 92 of airy_waves.py):
https://github.com/mancellin/capytaine/blob/f94afae9d1a56d9065cb67118134ab444ac7c1ee/capytaine/bem/airy_waves.py#L92
Is the minus there due to the force per unit area being minus pressure times the normal vector, or something else I'm missing?
Thank you very much as always!

[haskind.txt](https://github.com/mancellin/capytaine/files/7216126/haskind.txt)
(I had to rename the file to .txt to upload it)

mancellin:
Yes, that would be a nice example to add to the cookbook.

>  Is the minus there due to the force per unit area being minus pressure times the normal vector, or something else I'm missing?

Could be because the normal vector is the *outward* normal vector, but the force is the force *on* the body. Then it would be clearer to have the minus sign later in the function...

mancellin:
@marcogambarini Did you finally manage to figure out the reason for the discrepancy in the sign of the force?
Would you like to submit a PR adding the Haskind relation example to the cookbook?

mancellin:
> I can make a pull request for the cookbook, and also one to invert the signs in the function, if you want.

Yes, if it makes the code clearer, please submit a PR with the change of sign.

PR #129, opened 2022-03-14
Haskind

marcogambarini:
Cookbook example in which Haskind's relation is used to compute the excitation force from the radiation potential (see issue #100).


marcogambarini:
Ok, I added the line!

Issue #65, opened 2021-04-21
Parallel computation


freakingwave:
I have a 64-core workstation at hand. When I run Capytaine it only takes less than 10% of CPU. 
How can I deploy all CPU cores to run a task?

When I run it on a 8-core PC, it automatically takes 100% of the CPU.

Thanks

mancellin:
The number of threads is controlled by the environment variables `OMP_NUM_THREADS` and `MKL_NUM_THREADS`. Could you check their values?

freakingwave:
> 
> 
> The number of threads is controlled by the environment variables `OMP_NUM_THREADS` and `MKL_NUM_THREADS`. Could you check their values?

Thank you for reply.

by setting:
import os
 os.environ['OMP_NUM_THREADS']='32'
 os.environ['MKL_NUM_THREADS']='32'
 os.environ['NUMEXPR_MAX_THREADS'] = '32'

It still only takes less than 10% CPU. Time costs have no difference with setting 1, 8, 16, respectively.

Can you give some insights?

mancellin:
Changing the value from within Python seems to be buggy (see also #47).

If it is not the issue, could you tell me more about your OS and the version of Capytaine?

freakingwave:
I see...
I am using the latest 1.2.1 Capytaine by Spyder on Windows 10.

mancellin:
You should look on the internet for a way to change these variables system-wide on Windows 10 (sorry I can't help you with that).

Could you also check if it behaves the same with the previous version? (`conda install -c conda-forge capytaine=1.2.0`)

freakingwave:
Thanks.

1. Capy 1.2.0 behaves the same.
2. I set the global variables on Win10. It is weird. When I set threads to 1, it is obvious that only one thread is working. When I set them to a large number of threads like 16, 32 or 64, the CPU useage (about 12~15%) and cost time are the same.

mancellin:
I suppose it is not a bug, just the usual problem of diminishing returns in HPC...
I don't remember myself testing Capytaine with more than 8 cores. Maybe adding more threads does not improve the efficiency after some point. Or maybe your test problem is too simple to be worth using all the cores (the parallelism is on the number of panels in the mesh).

You could still maybe use all of your available power by setting OMP_NUM_THREADS to 4 or 8 and by launching at the same time several python scripts with different ranges of frequencies (for instance). But, you'll have to do it manually, there is no automated tools for that at the moment.

[...]

PR #136, opened 2022-05-02
Batch resolution in parallel with Joblib

### LINKED ISSUE #84: Problems and results should be picklable for parallelisation with Joblib
The library [Joblib](https://joblib.readthedocs.io/en/latest/) could be used to solve BEM problems in parallel, by letting the users replace the line (from the first example in the cookbook):
```
results = [solver.solve(pb) for pb in sorted(problems)]
```
by
```
from joblib import Parallel, delayed
results = Parallel(n_jobs=4)(delayed(solver.solve)(pb) for pb in sorted(problems))
```

However, it does not work right now because the `LinearPotentialFlowProblem` and `LinearPotentialFlowResult` objects (input and output of `solve`) are too fancy and can't be pickled by Joblib.

In particular, the overloading trick of lines https://github.com/mancellin/capytaine/blob/6b8dc62e88a6059597873f08ede18950c616c8f7/capytaine/bem/problems_and_results.py#L311-L316 might be a problem and should be replaced by something more straightforward.

mancellin:
Another layer of parallelization using the `joblib` library as an optional dependency.

- [x] Fix #84
- [x] Implement parallel resolution with Joblib
- [x] Figure out how to test that?
- [x] Add a simple test to the pytest directory
- [x] Add a line to the Changelog

### DIFF
--- a/capytaine/bem/problems_and_results.py
+++ b/capytaine/bem/problems_and_results.py
@@ -120,7 +120,7 @@ class LinearPotentialFlowProblem:
                 "g": self.g}
 
     @staticmethod
-    def group_for_parallel_resolution(problems):
+    def _group_for_parallel_resolution(problems):
         """Given a list of problems, returns a list of groups of problems, such
         that each group should be executed in the same process to benefit from
         caching.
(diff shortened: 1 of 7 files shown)


mancellin:
Would solve #65 (probably)

mancellin:
> Figure out how to test that?

So my current thought is that the pytest test suite should only check that the code is running fine even when using joblib.
The actual speedup due to parallelization should be tested somewhere else.
Probably in the `capytaine-benchmark` repository that I will work on someday soon.

PR #140, opened 2022-05-09
Fix sign of dissipation in impedance matrix.

### LINKED ISSUE #102: What is proper frequency domain convention for capytaine: exp(iwt) or exp(-iwt)?
Hi Guys, 

I am currently performing a floating WEC simulation in the capytaine and trying to compare it with WAMIT results. The added mass, radiation damping, and excitation forces are in good agreement, 

But RAO is not satisfactory when I used the default RAO function in capytaine `capytaine.post_pro.rao`.  

I think this is due to the convention of the equation of motion in the `impedance` function 
https://github.com/mancellin/capytaine/blob/6290cc0378ec769fdaa3220cd11a772793bab93d/capytaine/post_pro/impedance.py#L15

```python
A = (-omega**2*(dataset['mass'] + dataset['added_mass'])
         + 1j*omega*dataset['radiation_damping']
         + dataset['hydrostatic_stiffness'])
```
But when I changed the above code to the following code, the results are in good agreement. 

```python
A = (-omega**2*(dataset['mass'] + dataset['added_mass'])
         - 1j*omega*dataset['radiation_damping']
         + dataset['hydrostatic_stiffness'])
```

I think this due to the frequency domain convention of the NEMOH i.e., `exp(-iwt)`  (as mentioned in [here](https://lheea.ec-nantes.fr/valorisation/logiciels-et-brevets/nemoh-general-notations-and-conventions)) different from WAMIT `exp(iwt)`  (as mentioned in [here](https://www.wamit.com/manualupdate/v74_manual.pdf) pg 3-1). 

Please mention whether I am right or wrong. 

https://github.com/mancellin/capytaine/blob/6290cc0378ec769fdaa3220cd11a772793bab93d/capytaine/post_pro/impedance.py#L58

mancellin:
Fix #102

### DIFF
--- a/capytaine/post_pro/impedance.py
+++ b/capytaine/post_pro/impedance.py
@@ -55,11 +55,11 @@ def impedance(dataset, dissipation=None, stiffness=None):
     omega = dataset.coords['omega']  # Range of frequencies in the dataset
 
     A = (-omega**2*(dataset['mass'] + dataset['added_mass'])
-         + 1j*omega*dataset['radiation_damping']
+         - 1j*omega*dataset['radiation_damping']
          + dataset['hydrostatic_stiffness'])
 
     if dissipation is not None:
-        A = A + 1j*omega*dissipation
+        A = A - 1j*omega*dissipation
 
     if stiffness is not None:
         A = A + stiffness
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+* Fix major bug in impedance matrix and RAO computation: the sign of the
+  dissipation matrix was wrong in previous versions (:issue:`102` and
+  :pull:`140`).
+
 * Method :code:`show_matplotlib` can now colour mesh faces based on a
   specified scalar field (e.g. pressure) (:pull:`122`).
 


ryancoe:
@mancellin - @cmichelenstrofer noticed that the clumsy person (me!) who originally wrote `impedance.py` defined it in terms of the position (i.e., `Z = force / position`). While this is most convenient for the RAO, it is more common to define the impedance in terms of velocity (`Z = force / velocity`). The difference of course is simply `1/(j * omega)`, but it would probably be best to have the `impedance` function return the _**velocity impedance.**_

mancellin:
> The difference of course is simply `1/(j * omega)`

Or is it `1/(-j * omega)`? It is not the best PR to introduce another sign error :)

> it would probably be best to have the impedance function return the velocity impedance

I'm not sure it is worth a breaking change.
Maybe we can 
* keep `impedance` as it is, except for a couple lines in its docstring,
* add two new functions `position_impedance` and `velocity_impedance`, the former being just a single call to `impedance`.

ryancoe:
Good points. Your call on what is best in terms of avoiding a breaking change. I think either of the options you mention can work. It is never explicitly defined in `theory.rst`, but it is actually sort of implied that you are using the position impedance.

PR #141, opened 2022-05-09
Improve UI regarding wave direction in radians

### LINKED ISSUE #120: Wave direction in radians
In Capytaine, the direction of the incoming waves has to be specified in radians, not in degrees like in other software such as Nemoh. This can be confusing for users and I've seen calculations done with wave direction of 90 rad or 180 rad (https://github.com/mancellin/capytaine/issues/110#issue-1046529366; https://github.com/mancellin/capytaine/issues/30#issuecomment-1001913612).

A few things could be done to improve that:
* When [importing the legacy Nemoh config file format](https://github.com/mancellin/capytaine/blob/cb22cd2b455217dcbbcc19abaea3b52a35b42f92/capytaine/io/legacy.py#L20) to set up Capytaine, users expect to give directions in degrees as in Nemoh. Currently, the code does not convert the degrees into radians.
* When [setting up a problem](https://github.com/mancellin/capytaine/blob/cb22cd2b455217dcbbcc19abaea3b52a35b42f92/capytaine/bem/problems_and_results.py#L209), a warning could be displayed if the wave direction is not in the [-2π, 4π] range, since it may mean that the user gave a value in degrees.
* Somehow better highlight this in the documentation.

mancellin:
Fix #120

### DIFF
--- a/capytaine/bem/problems_and_results.py
+++ b/capytaine/bem/problems_and_results.py
@@ -222,6 +222,11 @@ class DiffractionProblem(LinearPotentialFlowProblem):
         super().__init__(body=body, free_surface=free_surface, sea_bottom=sea_bottom,
                          omega=omega, rho=rho, g=g)
 
+        if not (-2*np.pi-1e-3 <= self.wave_direction <= 2*np.pi+1e-3):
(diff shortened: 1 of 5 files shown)


mancellin:
@izabala123 FYI, the next version of Capytaine (probably labelled 1.4) will interpret the angles in Nemoh.cal as angles in degrees and not in radians.

PR #142, opened 2022-05-10
Add velocity impedance

mancellin:
Following the discussion in #140


mancellin:
Maybe, whenever we release a v2.0, we can switch for `impedance` to mean `velocity_impedance` by default. In the mean time, we should recommend that users explicitly choose between the explicit function names `velocity_impedance` or `position_impedance`.

Issue #50, opened 2021-03-15
Division by zero in OWC calculation


izabala123:
Dear Matthieu

I wanted to ask you advice to model an OWC. When modelling enclosed geometry this error message appears: 
Error: Impossible to compute the wave part of the Green function (division by zero).
I am using Capytaine loading nemoh.cal.

![image](https://user-images.githubusercontent.com/38589221/111211823-29be6480-85cf-11eb-9008-f7f5f784b083.png)

![image](https://user-images.githubusercontent.com/38589221/111211862-3642bd00-85cf-11eb-9c74-0770b96f8d08.png)

Thank you!

mancellin:
The problem comes probably from horizontal panels on the free surface (there is a division by z = 0).
I have on my todo list to provide a more helpful error message in this case.

izabala123:
I understand. Is there a way to solve it, maybe meshing differently?

mancellin:
For now: removing the panels on the free surface (the clipper of `body.keep_immersed_part()` does it IIRC) or sinking a little bit the mesh (`body.translate_z(...)`).
In the long run, it might be possible to implement a better handling of this singularity in the code (probably related to the singularity mentioned by @YingyiLiu in https://github.com/mancellin/capytaine/issues/44#issuecomment-797211251 about irregular frequencies removal).

izabala123:
Good idea. Thank you. 
I have tried it, but I get the next output:
```
INFO:   New floating body: .\mesh\oel.dat.
INFO:   New floating body: .\mesh\oel_lid.dat.
INFO:   New floating body: .\mesh\oel.dat+.\mesh\oel_lid.dat.
WARNING:        The mesh of the body .\mesh\oel.dat+.\mesh\oel_lid.dat is not inside the domain.
Check the position of the free_surface and the sea_bottom
or use body.keep_immersed_part() to clip the mesh.
```
After these messages, it appears:
```
WARNING:        No convergence of the GMRES. Error code: 23970
```

mancellin:
What have you tried exactly? The error message says you have panels above the free surface (or below the sea bottom).

izabala123:
Oh true, my fault!. Now fixed. Sorry.
Unfortunately, now the error is:
```
WARNING:        No convergence of the GMRES. Error code: 23970
```
The .tec files are empty.

mancellin:
[...]

PR #143, opened 2022-05-10
Improve error messages

mancellin:
Follow-up on #50

- [x] Division by zero
- [x] No-convergence of GMRES
- [x] Direct solver might fail without warning when there are overlapping panels


mancellin:
For overlapping panels, the error seems to be actually due to a NaN coming from the Green function Fortran code.
It is now caught earlier.

PR #149, opened 2022-05-23
Rename impedance

### LINKED ISSUE #147: position impedance -> (new name)
@cmichelenstrofer @dtgaebe @gbacelli and I were talking about velocity impedance and the RAO within our group (see #140 and also #142). @gbacelli pointed out that impedance should always be the ratio of effort over flow and that power is the product of these two.

```
Impedance: Z = effort / flow

Power: P = effort * flow
```

To avoid violating this convention, I would suggest that we 

1. rename `position_impedance` to something new (one suggestion is `pto_rao`) -- or remove this

https://github.com/mancellin/capytaine/blob/5c5305743b9759643f4a35c75e42e1207bf6e0b6/capytaine/post_pro/impedance.py#L74-L76

2. define the "_base_" impedance function for velocity impedance

https://github.com/mancellin/capytaine/blob/5c5305743b9759643f4a35c75e42e1207bf6e0b6/capytaine/post_pro/impedance.py#L15-L71

mancellin:
Fix #147

### DIFF
--- a/capytaine/post_pro/__init__.py
+++ b/capytaine/post_pro/__init__.py
@@ -2,5 +2,5 @@
 # See LICENSE file at <https://github.com/mancellin/capytaine>
 
 from capytaine.post_pro.rao import rao
-from capytaine.post_pro.impedance import impedance, position_impedance, velocity_impedance
+from capytaine.post_pro.impedance import impedance, rao_transfer_function
 from capytaine.post_pro.kochin import compute_kochin
--- a/capytaine/post_pro/impedance.py
+++ b/capytaine/post_pro/impedance.py
@@ -12,12 +12,53 @@ import xarray as xr
 LOG = logging.getLogger(__name__)
 
 
-def impedance(dataset, dissipation=None, stiffness=None):
-    """Complex-valued mechanical impedance matrix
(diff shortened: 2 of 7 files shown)


cmichelenstrofer:
@mancellin what is the motivation for using the $e^{-i\omega t}$ instead of positive $e^{i\omega t}$?

mancellin:
> @mancellin what is the motivation for using the e−iωt instead of positive eiωt?

Legacy.
Capytaine inherits it from Nemoh, which inherits it from Aquadyn, which might be follwoing other works in the 80's or earlier.
I can't really tell when and why this convention has been chosen.

mancellin:
It appears in a few other places, such as the relation between the pressure and the potential. It might be doable, but I don't want to do that hastily.

cmichelenstrofer:
> It appears in a few other places, such as the relation between the pressure and the potential

I see. It seems like this doesn't make a difference in the output results (e.g. diffraction, FK coefficients) but would need to be changed for consistency throughout the code. 

> but I'm not sure of all the consequences and I don't want to do that hastily

Makes sense. It shouldn't be part of this PR then. I will start a new issue to discuss if options.

PR #150, opened 2022-05-23
Fix and refactor Sphere class

mancellin:
While drafting some documentation for the `geometric_center` of bodies, I found a bug in the Sphere class. I took the opportunity to clean up the whole class and add some tests.


mancellin:
Stumbled upon #151...

Issue #38, opened 2020-07-10
Issue with deep panels and/or high frequencies


mancellin:
There might be a serious issue with the asymptotic formula for the Green function below 8λ (where λ is the wavelength of the waves). That is for deep panels and/or low wavelengths.

This is probably one of the reasons of the discrepancies in #20. The issue might already be present in Nemoh but got worse in Capytaine.

Thanks to @dav-og and his colleagues for pointing me in this direction. I'm looking for a fix.

mancellin:
> deeper than 8λ under the free surface

Dear @mancellin from 2020,

I know this has been a troubled year, but it is no excuse to confuse the wavenumber with the angular wavenumber (#160). Actually the issue is from panels deeper than 8λ/2π ≃ 1.2λ.

You~~rs~~,
@mancellin from 2022

Issue #20, opened 2019-10-10
Merge Nemoh's commit 115 in Capytaine


izabala123:
Dear Matthieu

I have had the opportunity of comparing Capytaine with Wamit and Nemoh version 115, for the DeepCWind wind turbine floating semisub platform ([https://community.ieawind.org/HigherLogic/System/DownloadDocumentFile.ashx?DocumentFileKey=12510238-c40b-9597-a845-580276c1b88d](https://community.ieawind.org/HigherLogic/System/DownloadDocumentFile.ashx?DocumentFileKey=12510238-c40b-9597-a845-580276c1b88d)) and it seems that the difference in results worth to include Nemoh irregular frequencies removal technology into Capytaine.
![View of the platform](https://github.com/izabala123/BEMRosetta/blob/6498c76bc694d4809938f0eab380c7ce0482684f/other/md%20resources/Other/tempdcw.png)
![Added mass in heave comparison from BEMRosetta](https://github.com/izabala123/BEMRosetta/blob/6498c76bc694d4809938f0eab380c7ce0482684f/other/md%20resources/Other/tempheave.jpg).
The legend means:
- "Capyta": Capytaine
- "Joined files": Nemoh v115
- "dcw10": Wamit

I can submit all the files necessary to reproduce this case test.

All the best
Iñaki Zabala

izabala123:
Of course yes. However I will need some days to prepare the binary and run the simulation.
In parallel, could you give me the files lacked in revision 118?

izabala123:
Dear Matthieu

These are the results of Nemoh  v113, Capytaine and Nemoh v115. 
It would be great if Capytaine could include v115 improvements. 

![Radiation damping 1 heave 3](https://user-images.githubusercontent.com/38589221/66680152-230c1f00-ec70-11e9-8de0-872238bab336.jpg)
![Added mass 1 heave 3](https://user-images.githubusercontent.com/38589221/66680169-2b645a00-ec70-11e9-9de7-a0cf8d8e69e4.jpg)

All the best
Iñaki Zabala

mancellin:
Thank you very much. There seems to be a significant discrepancy between Nemoh and Capytaine on this test case. I'll need to investigate that...

I have uploaded the version you've asked for: https://ancell.in/tmp/2019-10-11-Nemoh_UCD_89518c8a6edc10e601437fbe88b11da3.zip (I haven't run it for years, so I can't guarantee it will work fine...).

izabala123:
[...]

PR #156, opened 2022-06-16
Improve asymptotic behavior for high frequencies

mancellin:
This PR would fix one of the issues with high frequencies (cf. #20 and #38).

Wrong_asymptotics
(Dashed line is the current master branch. The plain line is the current pull request.)

The wrong asymptotic for high frequencies was caused by the following lines
https://github.com/mancellin/capytaine/blob/416175d102b9eb6f54d9d0f6775634afaddbfad4/capytaine/green_functions/Delhommeau_f90/Green_wave.f90#L186-L189
which were an attempt at transposing [these lines](https://github.com/LHEEA/Nemoh/blob/49393be96bd590e267bfbc16347b665cda0edeb8/Solver/Core/COMPUTE_GREEN_INFD.f90#L331-L336) and [these lines from Nemoh](https://github.com/LHEEA/Nemoh/blob/49393be96bd590e267bfbc16347b665cda0edeb8/Solver/Core/COMPUTE_GREEN_INFD.f90#L361-L366). I might have misunderstood the code from Nemoh, leading to a very different result.
Anyway, this if-else branch is not mentioned in any theoretical documentation and does not seem really justified nor necessary, so I'm just removing it in this PR.

Note that it does not solve all the issues with high frequencies. On the test case of #20, it seems to bring Capytaine close to Nemoh v113, but not to Nemoh v115. The remaining precision improvement will be handled in another PR.

### DIFF
--- a/capytaine/green_functions/Delhommeau_f90/Green_wave.f90
+++ b/capytaine/green_functions/Delhommeau_f90/Green_wave.f90
@@ -117,9 +117,9 @@ CONTAINS
       ERROR STOP
     ENDIF
 
-    !=====================================================================================
-    ! Evaluate the elementary integrals PDnX and PDnZ depending on dimless_Z and dimless_r
-    !=====================================================================================
+    !=======================================================================
+    ! Evaluate the elementary integrals depending on dimless_Z and dimless_r
+    !=======================================================================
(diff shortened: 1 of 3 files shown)


mancellin:
Also the PR fixes one of the tests in which the finite depth Green function was evaluated below the sea bottom.

Issue #115, opened 2021-11-25
Use scikit-build for PyPI package?


cmichelenstrofer:
I have not been able to install capytaine from PyPI using `pip` on my mac (default Fortran compiler does not work and haven't figured out a workaround). It installs with `conda` seamlessly. 

It seems like `scikit-build` would do for `pip` many of the things that `conda` does in the background to make Fortran compilation easier. [See this article.](https://python-packaging-tutorial.readthedocs.io/en/latest/binaries_dependencies.html) 
It might be worth exploring building the PyPI package with [scikit-build](https://scikit-build.readthedocs.io/en/latest/).

mancellin:
I agree that it would be nice to improve the `pip` packaging of Capytaine.
Thank you for the reference, I was not aware of `scikit-build`, I'll look into it.

cmichelenstrofer:
FYI for those of us using Macs:
 
Installing Capytaine requires compiling fortran. When installing with `pip` you need to make sure you have fortran compilers. **_Macs do not come with a Fortran compiler_**. 

**Install gfortran** 
`gfortran` is part of the GNU compiler Collection `gcc`. There are several ways to install it, here is how I did it using MacPorts:

1. install xcode (large) from Apple
2. install command-line-tools for xcode from Apple
3. install MacPorts
4. Install & Activate gcc (and gfortran) using MacPorts
```bash
sudo port install gcc11 +gfortran
sudo port select --set gcc mp-gcc11
```
5. Add ``export CC="gcc"`` to your `.zshrc` file

kmruehl:
@mancellin I"m also having issues with the FORTRAN compiler, but on Windows. I'm trying to pip install from GitHub, and am getting a `no Fortran compiler found` error.

mancellin:
There used to be [instructions in the doc on how to set up a fortran compiler on Windows](https://github.com/mancellin/capytaine/blob/v1.1/docs/developer_manual/installation.rst). But this is a bit cumbersome and I now recommend installing the precompiled binaries from `conda` for users and using Windows Subsystem for Linux for developpers.
Having precompiled binaries also on `pip` is on my roadmap, but it is a fair amount of work (and not the fun kind of work).

kmruehl:
@mancellin thanks. I was able to do a conda installation on Windows. A pip install from GitHub is my preference, so thank you for pointing me to the prior developer installation instructions (I couldn't find them).

mancellin:
[...]

PR #170, opened 2022-07-05
Fix installation of Capytaine with pip when Numpy is missing

### LINKED ISSUE #112: PyPI package broken
Currently `pip install capytaine` does not work unless you already have installed `numpy`. The problem is that although `numpy` is listed under `install_requires` in `setup.py`, it is used in that same file (`from numpy.distutils.core import Extension, setup`). 

I think the solution from the [second answer here](https://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py) should work. 

It would look something like:
```python
from setuptools import dist
dist.Distribution().fetch_build_eggs(['numpy'])
```

mancellin:
Fix #112, thanks to @cmichelenstrofer.
Also add a link to the installation instructions for macOS https://github.com/mancellin/capytaine/issues/115#issuecomment-1143987636.

### DIFF
--- a/docs/user_manual/installation.rst
+++ b/docs/user_manual/installation.rst
@@ -53,4 +53,5 @@ If you do, you can install Capytaine as::
 If you can't install a compiler, it is recommended to use Conda instead.
 
 .. [#] For example, on Ubuntu or Debian: :code:`sudo apt install gfortran`.
+       On macOS, see `for instance these instructions <https://github.com/mancellin/capytaine/issues/115#issuecomment-1143987636>`_.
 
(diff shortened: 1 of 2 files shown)


cmichelenstrofer:
@mancellin I think [Scikit-build](https://scikit-build.readthedocs.io/en/latest/) is still worth exploring so that Mac users don't have to manually install Fortran compilers and the process of installing with `pip` would hopefully be as easy as with `conda`.

mancellin:
@cmichelenstrofer I agree, but this is a larger endeavor that I keep for later.

PR #180, opened 2022-07-21
Improve XieDelhommeau implementation and documentation

mancellin:
Since version 1.1, there are two implementations of the Green function available in Capytaine. The main one is `Delhommeau`, which is basically the same as in Nemoh. The other one is `XieDelhommeau`, which is a slight variant.
Their results are usually almost identical. However, I noticed that I wasn't making full use of the change introduced by `XieDelhommeau`.

In this PR, the implementation of `XieDelhommeau` has been changed to exploit the higher precision method for the singular parts of the Green function.

For most cases, it makes no visible difference. For some cases in infinite depth with horizontal panels close to the free surface, it leads to a large gain in accuracy, at no computational cost.
The example below is the result of the new `XieDelhommeau` for a sphere of radius 1 and center (0, 0, -1.01). The coarse mesh leads to results much closer to the fine mesh when the new `XieDelhommeau` method is used. The former `XieDelhommeau` was basically overlapped with `Delhommeau`.
Example of result

Beside, another consequence of `XieDelhommeau` that wasn't exploited until now is a slight simplification of the code in finite depth (and maybe a marginal performance gain).

This PR also includes a revamp of the corresponding section of the documentation. The`XieDelhommeau` variant is actually slightly simpler to document and could become the base method in the future.

### DIFF
--- a/capytaine/green_functions/Delhommeau_f90/Green_wave.f90
+++ b/capytaine/green_functions/Delhommeau_f90/Green_wave.f90
@@ -73,9 +73,10 @@ CONTAINS
       !================================================
 
 #ifdef XIE_CORRECTION
-      VS(3) = CMPLX(integrals(1, 2)/PI - ONE/r1, integrals(2, 2), KIND=PRE)
+      VS(3) = CMPLX(integrals(1, 2)/PI + ONE/r1, integrals(2, 2), KIND=PRE)
 #else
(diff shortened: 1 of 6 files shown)


mancellin:
This PR will be closed and split into two: the implementation of the modified Green function will be moved to libDelhommeau whereas the documentation (some of it unrelated to XieDelhommeau) will be handled in another PR in Capytaine.

PR #181, opened 2022-07-26
Document the finite depth Green function

mancellin:


mancellin:
Will be done in libDelhommeau instead.

PR #183, opened 2022-07-27
remove volume/area methods from pre-defined bodies.

### LINKED ISSUE #182: Rectangle's volume after keep_immersed_part
```python
rect = cpy.RectangularParallelepiped()
rect.volume == rect.keep_immersed_part().volume
```

The volume should be halved but it isn't. This is becuase the volume is calculated from `self.size` ([line 296 in capytaine/bodies/predefined/rectangles.py](https://github.com/capytaine/capytaine/blob/master/capytaine/bodies/predefined/rectangles.py#L296)). 

Maybe the fix is as simple as removing the specialized version o f the `volume` method. But I don't know enough about these classes to tell if something else should be changed.

cmichelenstrofer:
fixes #182

### DIFF
--- a/capytaine/bodies/predefined/cylinders.py
+++ b/capytaine/bodies/predefined/cylinders.py
@@ -253,10 +253,6 @@ class HorizontalCylinder(FloatingBody):
 
             return cylinder
 
-    @property
-    def volume(self):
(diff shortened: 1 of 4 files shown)


mancellin:
Thank you!

PR #187, opened 2022-07-29
Reflect mesh across x and y axis according to flags in gdf

### LINKED ISSUE #186: Mesh loader not recognizing symmetry in gdf's
I am loading gdf files into Capytaine, but having issues due to the symmetry as defined in the header of the gdf. An example header is shown below with the 1's for ISX and ISY indicating symmetry about the x and y axes. 
image

The mesh_loaders.py script reads the flags for isx and isy in, but does not currently use the flags to reflect the mesh to create the full mesh.

jtgrasb:
This PR resolves #186.

### DIFF
--- a/capytaine/io/mesh_loaders.py
+++ b/capytaine/io/mesh_loaders.py
@@ -739,7 +739,14 @@ def load_GDF(filename, name=None):
 
     ifile.close()
 
-    return Mesh(vertices, faces, name)
+    if isx == '1' and isy == '1':
+        return ReflectionSymmetricMesh(ReflectionSymmetricMesh(Mesh(vertices, faces, f"quarter_of_{name}"), yOz_Plane, f"half_of_{name}"), xOz_Plane, name)
+    elif isx == '1':
+        return ReflectionSymmetricMesh(Mesh(vertices, faces, f"half_of_{name}"), xOz_Plane, name)
+    elif isy == '1':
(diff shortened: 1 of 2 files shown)


mancellin:
Looks good, thank you!
Would you mind adding an entry in the `docs/changelog.rst`?

jtgrasb:
@mancellin Added!

capytaine/io/mesh_loaders.py
```diff
@@ -739,7 +739,14 @@ def load_GDF(filename, name=None):
 
     ifile.close()
 
-    return Mesh(vertices, faces, name)
+    if isx == '1' and isy == '1':
+        return ReflectionSymmetricMesh(ReflectionSymmetricMesh(Mesh(vertices, faces, f"quarter_of_{name}"), yOz_Plane, f"half_of_{name}"), xOz_Plane, name)
+    elif isx == '1':
+        return ReflectionSymmetricMesh(Mesh(vertices, faces, f"half_of_{name}"), xOz_Plane, name)
```
mancellin, line 745: According to WAMIT's manual, `ISX = 1` means that the $x=0$ plane is a plane of symmetry. The $x=0$ plane would be `yOz_Plane` in Capytaine. Could you double check with an example?

jtgrasb: Yes, that is correct. Thanks for catching that. I only have an example that has both axes of symmetry, but the implementation at least functions for that example.

Issue #178, opened 2022-07-21
Potential issue with stiffness matrix (K44 and K55)


sdillenburg:
Dear all,

The horizontal (x and y) component of the center of gravity refers to the (local) coordinates of the center of the body, the vertical component, however, is defined relative to the global z = 0. See following example:

The function `FloatingBody.center_of_buoyancy ` returns the center of buoyancy in global coordinates. Following line suggests the center of gravity is also defined in global coordinates:

https://github.com/mancellin/capytaine/blob/8a07364afa8092cc30f1aefed57fab16b5c96034/docs/user_manual/examples/boat_animation.py#L18

However, when a body has its center off x = 0 and y = 0, results become strange, see following test:

```
import capytaine as cpt
import numpy as np
import xarray as xr
import matplotlib.pyplot as plt

# Define the bodies
# Mass matrix
m = 4612500 # kg
radii = np.array([9,9,12]) # m

M = np.eye(6)*m
M[3:,3:] *= radii**2

# Bodies
barge_0 = cpt.RectangularParallelepiped(size=(30,30,5), resolution=(20,20,5), center=(0,-20,-2.5),top=False,name="barge_0")
barge_0.center_of_mass = (0,0,5)
barge_0.add_all_rigid_body_dofs()
barge_0.hydrostatic_stiffness = barge_0.compute_hydrostatic_stiffness(rho=1025)
barge_0.inertia_matrix = xr.DataArray(M, dims=['influenced_dof', 'radiating_dof'],
                                         coords={'influenced_dof': list(barge_0.dofs),'radiating_dof': list(barge_0.dofs)},
                                         name='inertia_matrix')
                                         
barge_1 = cpt.RectangularParallelepiped(size=(30,30,5), resolution=(20,20,5), center=(0,-20,-2.5),top=False,name="barge_1")
barge_1.center_of_mass = (0,-20,5)
barge_1.add_all_rigid_body_dofs()
barge_1.hydrostatic_stiffness = barge_1.compute_hydrostatic_stiffness(rho=1025)
barge_1.inertia_matrix = xr.DataArray(M, dims=['influenced_dof', 'radiating_dof'],
                                         coords={'influenced_dof': list(barge_1.dofs),'radiating_dof': list(barge_1.dofs)},
                                         name='inertia_matrix')

print(barge_0.center_of_buoyancy)
print(barge_1.center_of_buoyancy)

problems_0 = []
problems_1 = []

[...]

PR #196, opened 2022-08-26
Make hydrostatics invariant by horizontal translation

mancellin:
Goal: fixing #178


mancellin:
> * Do some more manipulations to deduce $(x_c, y_c, z_c)$ from the displacement field of the rotation dofs.

I finally tried this option. It works but is probably full of broken edge cases. I guess it is good enough for a quick 1.4.1 release. The data model for rigid body dofs needs some rethinking from the ground up anyway...

PR #199, opened 2022-09-01
Change dof ordering in assemble_regular_array()

### LINKED ISSUE #198: Ordering of dofs in composite FloatingBody
When combining two bodies with the `+` operator, the dofs are ordered one body after the other.
```python
import capytaine as cpt

body = cpt.Sphere()
body.add_all_rigid_body_dofs()

pair = body + body.translated((1.0, 0.0, 0.0), name="other_sphere")
print(pair.dofs.keys())
```
```
dict_keys(['sphere_0__Surge', 'sphere_0__Sway', 'sphere_0__Heave', 'sphere_0__Roll', 'sphere_0__Pitch', 'sphere_0__Yaw', 'other_sphere__Surge', 'other_sphere__Sway', 'other _sphere__Heave', 'other_sphere__Roll', 'other_sphere__Pitch', 'other_sphere__Yaw'])
```

But when using `assemble_regular_array()`, this does not seem to be the case and the dofs are intermingled.
```python
array = body.assemble_regular_array(2, (1, 2))
print(array.dofs.keys())
```
```
dict_keys(['0_0__Surge', '0_1__Surge', '0_0__Sway', '0_1__Sway', '0_0__Heave', '0_1__Heave', '0_0__Roll', '0_1__Roll', '0_0__Pitch', '0_1__Pitch', '0_0__Yaw', '0_1__Yaw'])
```

One of them should be changed for consistency.
The behavior of `+` is more convenient as it allows to define block hydrostatic matrices as discussed in https://github.com/capytaine/capytaine/issues/188#issuecomment-1234021728.
So the behavior of `assemble_regular_array()` should be changed.

mancellin:
Fix #198

### DIFF
--- a/capytaine/bodies/bodies.py
+++ b/capytaine/bodies/bodies.py
@@ -760,7 +760,7 @@ respective inertia coefficients are assigned as NaN.")
         -------
         FloatingBody
         """
-        bodies = (self.translated((i*distance, j*distance, 0), name=f"{i}_{j}") for i in range(nb_bodies[0]) for j in range(nb_bodies[1]))
+        bodies = (self.translated((i*distance, j*distance, 0), name=f"{i}_{j}") for j in range(nb_bodies[1]) for i in range(nb_bodies[0]))
         array = FloatingBody.join_bodies(*bodies)
         array.mesh = build_regular_array_of_meshes(self.mesh, distance, nb_bodies)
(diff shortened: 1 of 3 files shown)


mancellin:
The commit ac034b8 fixed the issue, but maybe it would a good opportunity to remove duplicate code and use `FloatingBody.combine_dofs` there.

PR #222, opened 2022-10-04
__rmatmul__ for low rank matrices

marcogambarini:
Implemented __rmatmul__ for low rank matrices: if v is a vector and M is a low rank matrix, vM can be computed as M.__rmatmul__(v).

### DIFF
--- a/capytaine/matrices/low_rank.py
+++ b/capytaine/matrices/low_rank.py
@@ -376,6 +376,14 @@ class LowRankMatrix:
     def _mul_with_vector(self, other):
         return self.left_matrix @ (self.right_matrix @ other)
 
+    def __rmatmul__(self, other):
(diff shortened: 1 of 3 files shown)


mancellin:
Thank you! Looks good to me.
We should just add a small test in `pytest/test_matrices.py` and a line in the changelog.

mancellin:
Thank you for the last commits. Looks good, I'll merge that.

> Sorry - I made my usual mistake of working on the master branch, I will be more careful next time. I hope it's not too much of an issue!

Not an issue for me actually. I've chosen a development process in which I don't take your commits as they are, but I squash them together. It makes a cleaner git history, but also make the life of contributors like you a bit more difficult, since my master branch will diverge from yours.

PR #224, opened 2022-10-05
Change floating point precision of the Green function from Python

mancellin:
### Single precision versions

The Green functions now have a `floating_point_precision` argument that can be set to `float32` for single precision computation (default is still `float64`).

TODO: compare performance and memory usage of the single precision version.

#### Some comments on the implementation

The Fortran source code is compiled twice, once for each precision.

My first try to select a precision was to use the compiler's preprocessor:
```
#ifdef SINGLE_PRECISION
  integer, parameter :: pre = kind(1e0)
#else
  integer, parameter :: pre = kind(1d0)
#endif
  real(kind=pre) :: blablabla
```
It works well as a standalone, but f2py is not aware of the preprocessor and is always using the first definition of `pre`. Thus f2py is failing to infer the correct type signature for all functions and to bind the Fortran code to Python.

After some desperate efforst to configure f2py, I found another way that is still relatively straightforward: putting `pre` in its own module `floating_point_precision` and having two files `float32.f90` and `float64.f90` implementing two variants of the same module. Depending of the version of `floating_point_precision` that is included in the list of source files, the compiled code will be single or double precision. And f2py is now able to automatically infer the signatures of all functions.

### Untabulated Green function

Edit: I changed my mind and I will not include the untabulated Green function in this PR. There might be a nicer way to do that and this PR is big enough already.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ test: develop
 	python -m pytest
 
 clean:
-	rm -f capytaine/green_functions/*.so
+	rm -f capytaine/green_functions/libs/*.so
 	rm -rf build/
 	rm -rf dist/
 	rm -rf capytaine.egg-info/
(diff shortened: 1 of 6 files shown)


mancellin:
The effect on the computation time is rather small. An exemple on matrix building:
```
              --                                 depth
              -------------------------- ---------------------
               floating_point_precision      20.0       inf
              ========================== =========== =========
                       float32            514±0.7ms   368±1ms
                       float64             566±1ms    411±4ms
```

Issue #214, opened 2022-09-27
New declarative user interface for defining bodies


mancellin:
I'm really not satisfied with the current user interface for defining floating bodies.
It has been designed to be flexible and customizable, but I believe it makes defining simple cases (e.g. a single rigid body) too complicated.

### Current state

For instance, the "standard" way of defining a rigid body with hydrostatics looks like this:
```python
body = cpt.Sphere()
body.center_of_mass = (0, 0, -0.1)
body.add_all_rigid_body_dofs()
body.inertia_matrix = body.compute_rigid_body_inertia()
body.keep_immersed_part()
body.hydrostatic_stiffness = body.compute_hydrostatic_stiffness()
```

My main issue with this framework is that the ordering of the instructions can significantly change the output in a way that is not explicit or easy to predict.
For instance
```python
body.center_of_mass = (0, 0, -0.1)
body.add_all_rigid_body_dofs()
```
gives a different result than
```python
body.add_all_rigid_body_dofs()
body.center_of_mass = (0, 0, -0.1)
```
and users might not be aware of it.
See also #197, about the ordering of the instructions.

### Mockup of new design

One of the roots of the problem is the imperative style of coding that is used here.
Using a more declarative style for the user interface might be less error-prone.

```python
import capytaine as cpt

body = cpt.FloatingBody(
    mesh=cpt.mesh_sphere(radius=1.0, center=(0, 0, 0), resolution=(10, 20)),
    dofs=cpt.rigid_body_dofs(rotation_center=(0.0, 0.0, 1.0)),
    center_of_mass=(0.0, 0.0, -0.1),
)

problem = cpt.RadiationProblem(body=body.immersed_part(), omega=1.0, radiating_dof="Heave")

some_mesh = cpt.load_mesh("my_mesh.stl")
other_body = cpt.FloatingBody(
    mesh=some_mesh.translated([1.0, 0.0, 0.0]),
    dofs={'Bulge': [n for n in some_mesh.faces_normals]},
    hydrostatics=False,
)
```

Comments:
* The predefined bodies (`cpt.Sphere()`, etc.) return `FloatingBody` instances, although they actually only set the `mesh` attribute of the `FloatingBody`. New functions such as `mesh_sphere` returning only a mesh would make it more explicit that we are defining a new `FloatingBody` and leave the possibility to use the other arguments of `FloatingBody` constructor (`dofs`, etc.).
[...]

PR #231, opened 2022-10-12
New functions to load mesh

mancellin:
As part of #214, add functions `cpt.load_mesh` and `cpt.load_from_meshio` as alternative to `FloatingBody.from_file` and `FloatingBody.from_meshio`.


mancellin:
Forced pushed another commit instead because I had mistakenly commited the video output of `animate_boat.py`

Issue #228, opened 2022-10-12
Accuracy for thin plates


mancellin:
This issue investigates the results of the paper by Wanan Sheng et al (Ocean Engineering 2022) https://www.sciencedirect.com/science/article/pii/S0029801822003183

In this paper, the authors compare the results of Wamit, Hams and Nemoh on several meshes, including a cylinder with a thin damping plate.

![Mesh](https://user-images.githubusercontent.com/31126826/195331045-60c4368f-1395-49a5-8f2f-cc63a167a974.png)

Nemoh is found to be unreliable in this case, as seen in the graph below:

![Plot of the added mass from the aforementioned paper](https://user-images.githubusercontent.com/31126826/195330260-31d89b26-9c36-454f-a6b6-05cff54445be.png)

I tried to reproduce the above result (added mass of a heaving cylinder with a damping plate) with Capytaine.

In the graph below, the output of Capytaine for this mesh can be compared to the output of HAMS for the same mesh. The result of Capytaine seems to be the same as the result of Nemoh presented in the paper.

![Plot of the added mass from Capytaine](https://user-images.githubusercontent.com/31126826/195330254-218785fb-8031-41c6-9f3e-e7666f7a25f9.png)

When trying to reproduce the result by using my own mesh of the same geometry, the results were much closer to the ones of Wamit and Hams. It is unclear why the result is so sensitive to the mesh.

mancellin:
The paper implies that the inacurracy of Nemoh could be due to the use of a tabulation of the Green function. However, when disabling the tabulation in Capytaine (at the cost of a large increase of the computation time), no significant change is visible in the output.
![No tabulation](https://user-images.githubusercontent.com/31126826/195336753-57aebbf7-8292-4fee-bb9e-2673062578a0.png)

mancellin:
A possible explanation for the discrepancy between Nemoh/Capytaine and HAMS/Wamit is that the former use the *indirect* boundary integral equations while the latter use the *direct* boundary integral equations.
I have not been able to implement and validate the direct method in Capytaine yet, so I'm not really able to test this hypothesis at the moment.

mancellin:
This has mostly been fixed by #420 and #439. Detailed results are presented in an upcoming conference paper.

akeeste:
[...]

PR #234, opened 2022-10-19
Add another implementation of the Green function

mancellin:
Objective: offer users the possibility to do the computations with the implementation of [Liang, Wu and Noblesse](https://github.com/capytaine/LiangWuNoblesse_Green_function) of the Green function (available for infinite depth only). This is the method used in HAMS.

This is meant to help figuring out the cause of #228.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "capytaine/green_functions/libDelhommeau"]
 	path = capytaine/green_functions/libDelhommeau
 	url = https://github.com/capytaine/libDelhommeau
+[submodule "capytaine/green_functions/LiandWuNoblesse"]
+	path = capytaine/green_functions/LiandWuNoblesse
+	url = https://github.com/capytaine/LiangWuNoblesse_Green_function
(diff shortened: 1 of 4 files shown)


mancellin:
Close for now because of #284, but might make a new version of this later.

KapilKhanal:
@mancellin @cmichelenstrofer  For the derivative of this Green's function which is with respect to different variable, would it still work as this: D = K_rank + D_wave? where K_rank is the rankine integration code as implemented in Capytaine and D_wave is the normal derivative of the Green's function with respect to second point (at flowfield) as opposed to the first point (source point) in K_wave (based in delhommeau's approximation).

mancellin:
Do you ask whether

$$ \iint_\Gamma \nabla_1 G(x, \xi) \cdot n(x) d\xi = \iint_\Gamma \nabla_2 G(x, \xi) \cdot n(\xi) d\xi $$

for the Rankine Green function? They are not equal.

But they are related. Since #420, Capytaine can compute both, the difference is just a matter of changing a sign in the gradient and changing the normal vector.

PR #238, opened 2022-11-03
Update CI

mancellin:


mancellin:
Currently supported versions of Python: https://en.wikipedia.org/wiki/History_of_Python#Support

PR #241, opened 2022-11-10
Faster load_GDF function alternative.

ckesanapalli:
I created a new GDF function `load_GDF_compressed` that loads GDF files 4 times faster than the current `load_GDF` function.  

In addition to that, I also added the WAMIT parameter such `ulen, grav, isx, isy` as kwargs of `write_GDF` which can be assigned when creating a `.gdf` file.


mancellin:
Thanks for this PR!

I don't know if this part of the code is really performance critical, but your implementation clearly looks simpler and easier to maintain.

A few nitpicks:
* Maybe the test could be done on a mesh written by `write_gdf`. It would avoid adding the large `gdf` file to the repository, and allow to test the new parameters of `write_gdf`.
* The new parameters are missing in the docstring of `write_gdf`
* The call to `tuple` in `load_gdf_compressed` might not be necessary.

mancellin:
Also: not a requirement, but since `load_gdf` comes from [meshmagick](https://github.com/lheea/meshmagick), it might be fair-play to propose the patch also to them.

ckesanapalli:
> Also: not a requirement, but since `load_gdf` comes from [meshmagick](https://github.com/lheea/meshmagick), it might be fair-play to propose the patch also to them.

https://github.com/LHEEA/meshmagick/pull/34

pytest/test_io_meshes.py
```diff
@@ -290,3 +292,46 @@ def test_from_meshio_pygmsh(generate_pygmsh, tmp_path):
...
+        body_from_gdf.faces_radiuses, 
+        body_from_gdf_compressed.faces_radiuses
+        )
+    
+    np.testing.assert_allclose(
+        body_from_gdf.volume, 
+        body_from_gdf_compressed.volume
+        )
+    
+    body_path.unlink()
```
mancellin, line 336: FYI, there a feature in Pytest to create temporary directories, so you don't have to cleanup afterwards.
https://docs.pytest.org/en/6.2.x/tmpdir.html

capytaine/io/mesh_writers.py
```diff
@@ -474,6 +474,14 @@ def write_GDF(filename, vertices, faces):
...
     faces: ndarray
         numpy array of the faces' nodes connectivities
+    ulen: float, optional
+        length scale. The default is 100.0
+    gravity: float, optional
+        acceleration of gravity. The default is 9.81
+    isx: {0, 1}, optional
+        symmetry in x-axis. The default is 0
+    isy: {0, 1}, optional 
+        symmetry in y-axis. The default is 1
```
mancellin, line 484: I understand that you put 1 as default value to keep the previous behavior as default. But maybe it would make more sense to have 0 as default value.

capytaine/io/mesh_loaders.py
```diff
@@ -997,8 +1047,10 @@ def load_NEM(filename, name=None):
     'dat': load_MAR,
     'mar': load_MAR,
     'nemoh': load_MAR,
-    'wamit': load_GDF,
-    'gdf': load_GDF,
+    # 'wamit': load_GDF,
+    # 'gdf': load_GDF,
```
mancellin, line 1051: Maybe we could just replace the former `load_GDF`. Is there any good reason to keep it?

capytaine/io/mesh_writers.py
ckesanapalli: I totally agree with you.

mancellin:
Thank you @ckesanapalli 

Could you please add a line in the changelog?

I would have suggested to replace the old `load_gdf` by your implementation, but you're using it in the test, so maybe we can keep both.

PR #244, opened 2022-11-15
Add immersed_part method to floating bodies and meshes

mancellin:
This PR is part of #214.
The goal is to have less in-place transformations and to recommend to users a more functionnal approach to defining bodies.


mancellin:
Also the few references to the `body.full_body` parameter are being removed from the documentation (although the feature will stay until at least version 2.0).

PR #255, opened 2022-12-08
Included type casting for some user inputs

dav-og:
If a user defines these variable as integers then Capytaine writes them as integers when saving .nc files.

Including some type casting to make sure they are floats upon instantiation of the 'Problem' classes.

### DIFF
--- a/capytaine/bem/problems_and_results.py
+++ b/capytaine/bem/problems_and_results.py
@@ -28,7 +28,7 @@ class LinearPotentialFlowProblem:
     body: FloatingBody, optional
         The body interacting with the waves
     free_surface: float, optional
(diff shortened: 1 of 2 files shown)


dav-og:
@nhu-vnguyen, please can you check these instructions for compiling on Windows on your PC?

mancellin:
Thank you David!

I would have preferred separate PRs for the type casting and the doc. I would suggest creating a dedicated git branch and PR for each task instead of working on the master branch.

Could you add a line in `docs/changelog.rst` for the type casting issue? I'm not sure if it deserves to add a unit test to `pytest/`. You checked that it solves the issue with the nc file?

Issue #249, opened 2022-11-26
Add hydrostatic_stiffness matrix, to results dataset


davidfdomingos:
Good morning, I would like to leave a suggestion to add the hydrostatic_stiffness matrix to the results of  "cpt.BEMSolver().fill_dataset(test_matrix_radiation, body)".
I have noticed when trying to compute RAOs that the hydrostatic_stiffness matrix needs to be manually added to these results. Which according to the tutorials is not the case when solving radiation and diffraction problems separately.

mancellin:
Thank you for your comment.

There is something like that, if the body has an `hydrostatic_stiffness` parameter, it is added to the dataset:
```python
import xarray as xr
import capytaine as cpt

body = cpt.Sphere()
body.add_all_rigid_body_dofs()
body.center_of_mass = (0, 0, 0)
body.hydrostatic_stiffness = body.compute_hydrostatic_stiffness()
body.keep_immersed_part()

test_matrix = xr.Dataset(coords={"radiating_dof": ["Heave"], "omega": [1.0]})
ds = cpt.BEMSolver().fill_dataset(test_matrix, body)
print(ds["hydrostatic_stiffness"])
```
Is this what you have in mind?

Related maybe to #214, in which I'm trying to improve the way floating bodies are defined. In particular, I'm wondering if the hydrostatic stiffness should be automatically computed at the initialization of the body.

davidfdomingos:
Thanks! Indeed "body.hydrostatic_stiffness = body.compute_hydrostatic_stiffness()" works, I was trying before with "body.compute_hydrostatics()", but this didn't work.

mancellin:
Indeed, it would make sense for `body.compute_hydrostatic_stiffness()` to update the body to keep the matrices.

Issue #197, opened 2022-09-01
Should hydrostatics be computed with full mesh or clipped mesh?


mancellin:
> Edit: changed the code sample to define the center of mass of body before calling add_all_rigid_body_dofs(). In this way, the rotation dofs are defined around the center of mass and not the default (0, 0, 0).

So when I compute the inertia matrix before  calling `keep_immersed_part()`?Otherwise I can't get the inertia matrix for the whole rigid body.

_Originally posted by @ykshif in https://github.com/capytaine/capytaine/issues/188#issuecomment-1231794985_

mancellin:
Most hydrostatics magnitudes are computed on the clipped mesh (the `immersed_part()`).
For the inertia matrix, it would mean taking only the immersed mass into account, which is usually not what we want.
In `FloatingBody.compute_hydrostatics()`, the inertia matrix is computed on the clipped mesh, but I'm not sure it makes sense.

I guess, we should recommend to run the commands in the following order:
```python
body = cpt.Sphere()
body.center_of_mass = (0, 0, -0.1)
body.add_all_rigid_body_dofs()
body.inertia_matrix = body.compute_rigid_body_inertia()
body.keep_immersed_part()
body.hydrostatic_stiffness = body.compute_hydrostatic_stiffness()
```

I'm not satisfy by this user interface. It requires the users to order the operation in a way that can be hard to guess and to remember...

ykshif:
Hi mancellin,
When I calculate the inertia of Rectangular, I find that the full mesh and clipped mesh are not the same.Then I calculate the inertia of Sphere,the result is the same.I don't know if it's a bug?
`rectangle1 = cpt.RectangularParallelepiped(size=(1.0, 1.0, 1.0), resolution=(10, 10, 10), center=(0, 0, 0),name="body1Name")`
`rectangle1.center_of_mass = (0,0,0)`
`rectangle1.add_all_rigid_body_dofs()`
`rectangle1.inertia_matrix = rectangle1.compute_rigid_body_inertia()`
![image](https://user-images.githubusercontent.com/111233162/187878178-9d52daee-c804-450f-afa4-9f3827ada900.png)
When I clipped mesh.
![image](https://user-images.githubusercontent.com/111233162/187878400-7fc5aa77-f62a-477c-8fae-e0faa566e159.png)

sdillenburg:
[...]

PR #262, opened 2022-12-12
Fix compute_hydrostatics

### LINKED ISSUE #258: Fixing `body.compute_hydrostatics()`
I believe there are several issues with `body.compute_hydrostatics()`. I would like to make it more prominent in the tutorial and examples as part of #214, but that would requiring changing the following:

- The method runs `body.keep_immersed_part()` modifying the body, which might be unexpected by the user. As I'm trying to phase out in-place transformations in the high-level user interface in #214, it would make sense to avoid that.
- As noticed in #249, it would be more intuitive if the matrices computed with `compute_hydrostatics()` were stored with the body for further use.
- As mentionned in #197, it seems that the rigid body inertia matrix should be computed on the full mesh unlike the hydrostatic stiffness.

mancellin:
Closes #258


mancellin:
Also closes #197 and #249

mancellin:
Closes #197
Closes #249

PR #269, opened 2022-12-13
Check in CI that the package is correctly installed

mancellin:


mancellin:
I've had an issue today with the conda-forge build of the new release. I was wondering why the Github Action CI did not detect it.
The answer is that the CI is using `pip install -e .` which might not detect some installation issues occuring when using `pip install .`

PR #270, opened 2022-12-13
add long_name to dataset attrs

ryancoe:
When plotting directly from the `xr.Dataset` produced by `capytaine`, you can get nicer axis labelling by populating the `attrs['long_name']` entry. Units can also be added, but given the generalized modes I think it is probably best to leave this out.


mancellin:
Learned something, thank you!

Issue #273, opened 2022-12-21
pip install fails after NumPy 1.24


michaelcdevin:
Our [WEC optimization tool](https://github.com/SNL-WaterPower/WecOptTool) includes Capytaine as a dependency. When running the package's CI testing suite, Capytaine has begun to encounter an exception when compiling the Fortran binaries when installing from pip:
```
Run python3 -m pip install --upgrade pip
Requirement already satisfied: pip in /opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages (22.3.1)
Processing /home/runner/work/WecOptTool/WecOptTool
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Collecting meshio
  Downloading meshio-5.3.4-py3-none-any.whl (167 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 167.7/167.7 kB 3.1 MB/s eta 0:00:00
Collecting scipy
  Downloading scipy-1.9.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (33.7 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 33.7/33.7 MB 53.7 MB/s eta 0:00:00
Collecting capytaine>=1.4
  Downloading capytaine-1.5.post1.tar.gz (129 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 129.0/129.0 kB 33.4 MB/s eta 0:00:00
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [86 lines of output]
      /opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
        warnings.warn(
      WARNING: The wheel package is not available.
[...]

PR #274, opened 2022-12-22
Recast some user inputs to floats

dav-og:


mancellin:
Thank you David! I agree about the default for `wave_direction`.

It seems that the CI is hitting #273.

Issue #133, opened 2022-04-25
Removing the "convention" parameter


mancellin:
In Capytaine, several functions (e.g. `DiffractionProblem` or `froude_krylov_force`) have a `convention` parameter. 
Nemoh and WAMIT use different conventions for the incoming wave field and this parameter was meant to allow user to switch between them.

In an upcoming release, I plan remove this feature for the following reasons:
* Its original purpose was to simplify the comparison with results produced by other BEM solver. But other software such as [BEMIO](http://wec-sim.github.io/bemio/) or [BEMRosetta](https://github.com/BEMRosetta/BEMRosetta) already exist to do that kind of comparison. It makes sense to defer the handling of the different conventions to their import methods.
* It is barely documented.
* It is not really complete, since the frequency domain convention (cf. #102) cannot be changed.
* Since most tests use only the Nemoh convention, I'm not certain that the implementation is correct.
* Switching between different conventions within the same code can become insanely difficult to implement properly. Using a single convention in Capytaine would make Capytaine's developpers' life easier. It would also be much simpler for the downstream software importing results from Capytaine (such as the aforementionned BEMIO and BEMRosetta) if the convention used by Capytaine is always the same.

Thus, in my opinion this feature has a high cost for a small gain. In the future, all the internal routines of Capytaine should assume the Nemoh convention. Only import and export methods might handle other conventions.

If you have an opinion on this topic, feel free to leave a message below.

PR #281, opened 2023-01-23
Removing the convention parameter.

mancellin:
Adding the corresponding know-how in the documentation.


mancellin:
See #133 for motivation.

PR #285, opened 2023-01-26
add a function to write hydrostatics .dat files

dav-og:
Users can now use write Hydrostatics.dat and KH.dat files in the original Nemoh format using:

`capytaine.io.legacy.export_hydrostatics(hydrostatics_directory, bodies)`

- hydrostatics_directory has no default value
- bodies can be a single FloatingBody or a list (len >= 1) of FloatingBody objects

This PR improves BEMIO/WEC-Sim compatibility; there is now no need for custom user approaches to create these .dat files.

TODO: add examples/tests.


capytaine/io/legacy.py
```diff
@@ -259,3 +259,41 @@ def write_dataset_as_tecplot_files(results_directory, data):
...
+                line =  f'XF = {body.center_of_buoyancy[j]:7.3f} - XG = {body.center_of_mass[j]:7.3f} \n'
+                hf.write(line)
+            line = f'Displacement = {body.volume:E}'
+            hf.write(line)
+            hf.close()
+        np.savetxt(kh_file_path, body.hydrostatic_stiffness.values)
+
+    hydrostatics_file_name = "Hydrostatics.dat"
+    kh_file_name = "KH.dat"
+    try:
```
mancellin, line 284: Alternatively, you could do:
```python
if isinstance(bodies, FloatingBody) or len(bodies) == 1:
  ...  # Write single body
else:
  ...  # Write multiple body
```

mancellin:
Thank you @dav-og. Looks good to me. Will merge as soon as you add a few lines of documentation and a test.

capytaine/io/legacy.py
dav-og: I did try that but you still end up having to do either `bodies.center_of_buoyancy` or `bodies[0].center_of_buoyancy`?

mancellin: Good point.

Otherwise, wrap the single `FloatingBody` in a list at the beginning of the function:
```python
def export_hydrostatics(hydrostatics_directory, bodies):
  if isinstance(bodies, FloatingBody):
    bodies = [bodies]
  ... # continue as if the input was a list
```

Because the `except TypeError` does not make it very explicit that this case is for a single body outside of a list.

dav-og: yeah thats a fair point - I've changed it to do FloatingBody -> [FloatingBody] 👍

mancellin:
Maybe you can re-read the written file and then test on the numerical values using `np.testing.assert_allclose` to avoid comparing strings?

mancellin:
🎉

The documentation renders well. But I don't think the `compute_rigid_body_inertia` line in the example is necessary.

Could you also add a line to `docs/changelog.rst`?

PR #288, opened 2023-01-31
Return the full gradient of the Green function for velocity reconstruction

mancellin:
And not just the gradient in the normal direction.
For prototyping of velocity field reconstruction and direct solver.


mancellin:
The goal of the PR is to add a method `get_velocity_on_mesh()` to the solver to compute the velocity in the domain as a post-processing, similarly to the existing `get_potential_on_mesh()`.
This is a requirement for an on-going work on the implementation of forward speed.

PR #305, opened 2023-03-20
Clarify definitions in theory manual

mancellin:


mancellin:
FYI @nhu-vnguyen, that follows from our discussion last week.

PR #324, opened 2023-04-18
Distribute Capytaine with PyInstaller

mancellin:
Trying to build an independent executable from the conda-forge package.

TODO:
- [X] Have an executable working
- [x] Distribute the Nemoh-like command-line interface
- [ ] Distribute an executable able to run a Python script invoking Capytaine...
- [ ] ... including plots and 3d visualisation.
- [ ] Windows version
- [ ] MacOS version

### DIFF
--- a/run_pyinstaller.sh
+++ b/run_pyinstaller.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env sh
+
+# Run in a environment created with
+# conda create -n capy_install -c conda-forge capytaine pyinstaller
+
+CAPY_BINARIES_DIR="$CONDA_PREFIX/lib/python3.10/site-packages/capytaine/green_functions/libs/"
+
+BINARIES=""
+for f in $CAPY_BINARIES_DIR/*; do
+    BINARIES="--add-binary=$f:capytaine/green_functions/libs/ $BINARIES";
+done
+
+pyinstaller --clean --onefile --noconfirm $BINARIES examples/custom_dofs.py
--- a/run_pyinstaller.sh
+++ b/run_pyinstaller.sh
@@ -10,4 +10,4 @@ for f in $CAPY_BINARIES_DIR/*; do
     BINARIES="--add-binary=$f:capytaine/green_functions/libs/ $BINARIES";
 done
 
-pyinstaller --clean --onefile --noconfirm $BINARIES examples/custom_dofs.py
+pyinstaller --clean --onefile --noconfirm $BINARIES $CONDA_PREFIX/bin/capytaine


mancellin:
Thanks! 
I managed to build the executable also on Linux (editing the spec file is more portable than my experiments with a shell script...), but the matplotlib display does not seem to work. Is it working for you?

RubendeBruin:
matplotlib works fine here.

mancellin:
Never mind, I tried again in a clean conda environment created with `conda create -c conda-forge capytaine matplotlib pyinstaller` and it worked.

RubendeBruin:
Which imports (besides VTK) would we need to run the 3d visualizations?
Are there any example scripts/recipes that I can test on?

mancellin:
> Which imports (besides VTK) would we need to run the 3d visualizations? Are there any example scripts/recipes that I can test on?

Only VTK I believe.
In the example directory/cookbook, the `animate_free_surface.py` script and `boat_animation.py` script are using a lot of VTK. There is also a `body.show()` in `symmetric_body.py`.

RubendeBruin:
@mancellin ,

capytaine 2.0 seems to be installing just fine with with pip on windows11. No need for any compilers:
image

would this PR still be relevant?

mancellin:
@RubendeBruin Thank you for the feedback!
Yes, I'm still considering publishing a standalone version since many users might not want to deal with package management and virtual environments.

RubendeBruin:
Ok, in that case it may be easiest to use the embed() function of ipython to create a stand-alone ipython version with capytaine and vtk and all included.

Also, using a venv instead of a conda environment drastically reduces the size of the final executable. This has to do with numpy which is installed with a different math kernel if installed from pip.

mancellin:
Following up in its own repository: https://github.com/capytaine/capytaine-standalone

Still some issues with matplotlib at the moment (but VTK works :man_shrugging:).

mancellin:
The standalone distribution at https://github.com/capytaine/capytaine-standalone is (mostly) working.

ttammine:
Hello,

I took my first steps with Capytaine and tested the standalone distribution package.  It seems to be working with Windows  11 nicely but I think there is some problem with capytaine.io.legacy call. I got the following error:

In [39]: from capytaine.io.legacy import export_hydrostatics
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[39], line 1
----> 1 from capytaine.io.legacy import export_hydrostatics

ModuleNotFoundError: No module named 'capytaine.io.legacy'

mancellin:
Thank you @ttammine!
I can reproduce the issue. Not sure why some modules are absent from the standalone distribution. I'll look into it. I opened #448 to keep track of this.

Issue #88, opened 2021-08-10
Datasets and free_surface=infty


mancellin:
`cpt.assemble_dataset` badly handles the free surface z-coordinate parameter.

For instance, it fails with a cryptic error message when the data contains both cases with a free surface (`free_surface=0`) and without a free surface (`free_surface=np.infty`):
```python
#!/usr/bin/env python

import numpy as np
import capytaine as cpt

cylinder = cpt.HorizontalCylinder(
    length=10.0, radius=1.0, center=(0, 0, -2),
    nr=5, nx=40, ntheta=20, clever=False,
)
cylinder.add_all_rigid_body_dofs()
omega_range = np.linspace(0.1, 6.0, 2)

problems = [
    cpt.RadiationProblem(body=cylinder, sea_bottom=-np.infty, free_surface=fs, radiating_dof=dof, omega=omega)
    for fs in [0.0, np.infty]
    for dof in cylinder.dofs
    for omega in omega_range
]
solver = cpt.BEMSolver()
results = [solver.solve(pb) for pb in sorted(problems)]
data = cpt.assemble_dataset(results)
```
```
[...]

ValueError: cannot convert a DataFrame with a non-unique MultiIndex into xarray
```

Also, when a single value is given to `free_surface`, the resulting dataset does not keep this value neither as coordinate, nor as attribute.

This issue comes directly from another issue: having two different conventions for water depth.
* `LinearPotentialFlowProblem` uses `sea_bottom` and `free_surface` to store the z-coordinates of the boundaries of the domain.
* `assemble_dataset` and `fill_dataset` use `depth` to store the size of the domain.

I'm not sure what is the best way out of this issue.

The simplest would be to assume that the `free_surface=infty` case is just not supported in datasets and can be used only by manually defining `RadiationProblem`. It could make sense since `free_surface=infty` is mostly meant for teaching and testing and since datasets usually contain fields that do not make sense without a free surface (`omega`, `excitation_force`, ...).

mancellin:
Whenever we can do a big breaking change, we should maybe replace `free_surface` and `sea_bottom` in the `RadiationProblem` and `DiffractionProblem` by a single `water_depth` parameter. And create a new kind of problem for the edge case of resolution without a free surface.

mancellin:
I believe most of this issue has been solved by #89 and #340.

PR #340, opened 2023-05-17
Allow users to set `water_depth` instead of `sea_bottom`

mancellin:
This change is meant to uniformize notations in the code and replace the use of `sea_bottom` by `water_depth`.

Related to #88.


mancellin:
TODO: fix `keep_immersed_part` to allow `water_depth` as argument.

mancellin:
This (non-breaking) PR adds a `water_depth` alternative to user-facing `sea_bottom`.
Another future breaking PR might replace the usage of `sea_bottom` within the internals of the code.

PR #341, opened 2023-05-17
Add minimal_computable_wavelength property to bodies.

mancellin:
It is meant to be used as a lower bound for wavelength ranges, in order to avoid the "insufficient mesh resolution" warnings.

### DIFF
--- a/capytaine/bem/problems_and_results.py
+++ b/capytaine/bem/problems_and_results.py
@@ -129,7 +129,7 @@ class LinearPotentialFlowProblem:
                     "or use body.keep_immersed_part() to clip the mesh."
                 )
 
-            if self.wavelength < 8*self.body.mesh.faces_radiuses.max():
+            if self.wavelength < self.body.minimal_computable_wavelength:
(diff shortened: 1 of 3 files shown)


jtgrasb:
What is the theory behind this minimal wavelength? Can you direct me to a source where I can read up on it? 
--Thanks

mancellin:
The main theoretical justification is [Nyquist–Shannon sampling theorem](https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem), that say that you need at least 2 data points per wavelength to discretize a signal.
The value of 4 panels per wavelength is a rule of thumb coming from Nemoh's documentation.

In practice, the results can be surprisingly good for lower wavelengths, but I'm not aware of any deeper study of this topic.

mancellin:
I have in my notes a link to this paper, that I haven't had time to read yet: https://www.worldscientific.com/doi/abs/10.1142/S0218396X02001401

jtgrasb:
Thank you!

PR #329, opened 2023-04-21
Build wheels for Linux and MacOS with Github Actions

mancellin:

PR #342, opened 2023-05-19
Fix Windows wheel building

mancellin:
Based on the following line from Scipy's meson.build file:
```meson
# Adding at project level causes many spurious -lgfortran flags.
add_languages('fortran', native: false)
```
I'm testing if it helps with my Windows compile issues.

Edit: it did not. But explicitely linking `quadmath` fixed the issue from #329


mancellin:
:tada:

PR #344, opened 2023-05-31
Clipping - plane origin

bdeket:
I might be misunderstanding how clipping is supposed to work, but currently
`amesh.immersed_part(free_surface=1) == amesh.immersed_part(free_surface=-1)`

I tracked it down to the `Plane.get_Edge_Intersection` method, where I think the `dot`product should be used to calculate the intermediate point, instead of the `norm`

```
import numpy as np
import capytaine as cpt

myMesh = cpt.meshes.Mesh
myClip = cpt.meshes.clipper.clip
myPlane = cpt.meshes.geometry.Plane

a = myMesh([[0,-2,-2],[0,2,-2],[0,2,2],[0,-2,2]],[[0,1,2,3]])

def getEdgeIntersection(P,p0,p1):
    p0n = np.dot(p0, P.normal)
    p1n = np.dot(p1, P.normal)
   ## use dot-product instead of norm ##
    c = np.dot(P.normal, P.point)
    t = (p0n - c) / (p0n - p1n)
    if t < 0. or t > 1.:
        raise RuntimeError('Intersection is outside the edge')
    return (1-t) * p0 + t * p1
    
def testit(P):
    b = myClip(a, P)
    c = P.get_edge_intersection(a.vertices[1], a.vertices[2])
    d = getEdgeIntersection(P, a.vertices[1], a.vertices[2])
    print(b.vertices)
    print(b.faces)
    print(c, " =?= ", d)

print("below 1")
testit(myPlane(point=[0,0,1]))
# expected [[0,-2,-2],[0,2,-2],[0,2,1],[0,-2,1]]
# ok

print("below -1")
testit(myPlane(point=[0,0,-1]))
# expected [[0,-2,-2],[0,2,-2],[0,2,-1],[0,-2,-1]]
# NOT OK : gives below 1

print("above 1")
testit(myPlane(normal=[0,0,-1],point=[0,0,1]))
# expected [[0,-2,1],[0,2,1],[0,2,2],[0,-2,2]]
# NOT OK : gives above -1

print("above -1")
testit(myPlane(normal=[0,0,-1],point=[0,0,-1]))
# expected [[0,-2,-1],[0,2,-1],[0,2,-2],[0,-2,-2]]
# ok

print("below slopeR 1")
testit(myPlane(normal=[0,0.1,1],point=[0,0,1]))
# expected [[0,-2,-2],[0,2,-2],[0,2,0.8],[0,-2,1.2]]
# OK

print("below slopeL 1")
testit(myPlane(normal=[0,-0.1,1],point=[0,0,1]))
# expected [[0,-2,-2],[0,2,-2],[0,2,1.2],[0,-2,0.8]]
# ok

print("below slopeR -1")
testit(myPlane(normal=[0,0.1,1],point=[0,0,-1]))
# expected [[0,-2,-2],[0,2,-2],[0,2,-1.2],[0,-2,-0.8]]
# NOT OK

print("below slopeL -1")
testit(myPlane(normal=[0,-0.1,1],point=[0,0,-1]))
# expected [[0,-2,-2],[0,2,-2],[0,2,-0.8],[0,-2,-1.2]]
# NOT OK
```

### DIFF
--- a/capytaine/meshes/geometry.py
+++ b/capytaine/meshes/geometry.py
@@ -299,6 +299,11 @@ class Plane(Abstract3DObject):
         """Distance from plane to origin."""
         return np.linalg.norm(self.normal @ self.point)
 
+    @property
+    def s(self):
+        """Distance from origin to plane along the normal"""
+        return np.dot(self.normal, self.point)
+
     #################################
     #  Transformation of the plane  #
     #################################
@@ -365,7 +370,7 @@ class Plane(Abstract3DObject):
 
         p0n = np.dot(p0, self.normal)
         p1n = np.dot(p1, self.normal)
-        t = (p0n - self.c) / (p0n - p1n)
+        t = (p0n - self.s) / (p0n - p1n)
         if t < 0. or t > 1.:
             raise RuntimeError('Intersection is outside the edge')
         return (1-t) * p0 + t * p1
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -55,6 +55,8 @@ Bug fixes
 
 * Convert ``center_of_mass`` and ``rotation_center`` to arrays in :class:`~capytaine.bodies.bodies.FloatingBody` constructor to avoid a few issues (:issue:`319` and :pull:`325`).
 
+* Fix bug (leading to either ``RuntimeError`` or wrong output) when clipping with plane that do not contain the origin. (:pull:`344`)
+
 Internals
 ~~~~~~~~~
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -55,7 +55,7 @@ Bug fixes
 
 * Convert ``center_of_mass`` and ``rotation_center`` to arrays in :class:`~capytaine.bodies.bodies.FloatingBody` constructor to avoid a few issues (:issue:`319` and :pull:`325`).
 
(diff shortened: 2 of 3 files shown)


mancellin:
Good catch.

I could not reproduce a case where
> `amesh.immersed_part(free_surface=1) == amesh.immersed_part(free_surface=-1)`

but there is indeed something buggy around `get_edge_intersection`, as I get a runtime error when trying
```python
cpt.mesh_vertical_cylinder().clipped(cpt.Plane(point=(0, 0, -0.1), normal=(0, 0, 1)))
```

Would you mind adding a few tests in `pytest/test_meshes.py` based on your tries with your `testit` function?

bdeket:
```
cpt.meshes.predefined.mesh_parallelepiped(size=(5,5,5),resolution=(1,1,1)).immersed_part(-1) == cpt.meshes.predefined.mesh_parallelepiped(size=(5,5,5),resolution=(1,1,1)).immersed_part(1)
```
but in general rounding and vertices/faces order prevents it from being exactly true.

mancellin:
Thank you!

I fixed the failing test. `immersed_part` has `free_surface=0.0` argument by default.

mancellin:
Everything looks good to me. I've merged. Thanks again!

PR #353, opened 2023-06-13
Increase default tabulation size.

mancellin:
Make the improvements of https://github.com/capytaine/capytaine/pull/157#issuecomment-1167271580 the new default.


mancellin:
TODO: check that this does not increase too much the warm-up time.

PR #358, opened 2023-06-20
Investigate test failing in CI.

mancellin:


mancellin:
The failing test is probably an issue with floating point rounding errors. But the current approach to test mesh equality makes it difficult to fix.

PR #360, opened 2023-06-20
Exclude Fortran source files from wheel.

mancellin:
Waiting for Meson to fix bug to test that.


mancellin:
Don't remember which meson bug it was, but it seems to be working now.

Issue #369, opened 2023-07-17
Infinite Frequency Not Supported Error


Chandler192:
I have been running capytaine for several months on a windows computer running anaconda and I had the following line of code to calculate the added mass at infinite frequency:

infprob = [cpt.RadiationProblem(body=cpt_hydrobody, omega=np.inf, radiating_dof=dof) for dof in ['Surge', 'Sway', 'Heave', 'Roll', 'Pitch', 'Yaw']]

I then swapped to an ubuntu machine and reinstalled capytaine but now when I run the same code it gives me the following error:

Zero and infinite frequencies are currently not supported.

ANy help you can offer on this issue would be greatly appreciated.

mancellin:
That's my fault, sorry about that. The $\omega = 0$ and $\omega = \infty$ feature was conflicting with other new stuff in v2.0, so I disabled it and did have time to re-implement it cleanly before the release of v2.0.
It will be re-enabled in the next release in a few weeks.
In the mean time, if you need it, you can force `conda` to install the previous version with something like
```
conda create --name capy1.5 --channel conda-forge capytaine==1.5
conda activate capy1.5
```

PR #385, opened 2023-08-09
Reimplement zero and infinite frequencies

mancellin:
In order to avoid division by zero when computing something roughly like the following:
```python
b = 0 * a
c = b / 0
assert a == c
```
we define a special object that represent the zero that is simplified in the computation:
```python
zero = SymbolicMultiplication("0")
b = zero * a
c = b / zero
assert a == c
assert float(zero) == 0.0
```
so all the computations with `omega=zero` should work seamlessly. And similarly for infinity.


mancellin:
TODO: documentation and changelog

mancellin:
Fix #369

PR #391, opened 2023-09-13
Fix variable names in libDelhommeau

mancellin:
The term labelled "symmetric" and "anti-symmetric" were not actually what they claimed.


mancellin:
When updating the test, I took the opportunity to remove the dependency to hypothesis for testing. It was really not bringing a lot...

PR #404, opened 2023-10-10
Precompute the tabulation at compile time

mancellin:
Instead of building the Green function tabulation at each initialization of the BEMSolver, a tabulation could be stored in a file and loaded at initialization.

There are several options:
- Compute it at first initialization, store it in a cache directory and reuse it when required.
- Compute it at compile time, ship it in the precompiled package on PyPI or conda-forge. Only the default tabulation would be precomputed, as it is sufficient for 99% of users. It would approximately double the size of the package...

Right now, I'm investigating the second option, but maybe the first one is more pragmatic.


mancellin:
Gave up in favor of #454

PR #416, opened 2023-10-23
Quadratures without dependencies

mancellin:
The goal is to have higher-order integration scheme on some panels of the mesh sometimes (in particular, on the free surface for lid-based irregular frequencies removal).
Since Quadpy went commercial and closed-source, it cannot be required for the user to install it.
Instead, simple quadrature schemes can easily be implemented in Capytaine directly.


mancellin:
Left for another PR: different order of quadrature scheme depending on the panel.

PR #417, opened 2023-10-24
Fix computing quadrature before defining a body.

mancellin:
Seems to be a bug in the definition of quadrature schemes: schemes associated to a mesh before the body is defined are forgotten when the body is defined.


mancellin:
It is because of this line: https://github.com/capytaine/capytaine/blob/b8ada4407d8cba62d403e772e67cf73878060e2f/capytaine/bodies/bodies.py#L88
which reset the mesh quadrature.

PR #420, opened 2023-10-27
Direct method implementation

jnvn7:
@mancellin please review the direct method implementation. The option can be switched on with the 'direct_method = True'  as in the following:

BEMSolver().fill_dataset(test_matrix, body, direct_method=True)


mancellin:
Great, thank you! I'll test this.

mancellin:
I fixed the few comments I had.
I guess we are mostly missing documentation and some more testing now.

mancellin:
I was thinking whether we should rename the optional argument `direct_method=True`.

I can propose the following, that might slightly more future-proof:
- optional argument `method="direct"` or `method="indirect"` (default) in `capytaine/bem/solver.py`, since we could consider adding other boundary integral methods.
- a boolean to switch between the $D$ and $K$ matrices in `capytaine/bem/engines.py`, `capytaine/green_functions` and `libDelhommeau`. Other methods might use the same matrices in different ways, so it might be best if they are referred to by their name and not by one of the method they are employed with. I believe their mathematical names are respectively *double layer operator* and *adjoint double layer operator*, so the optional argument could be `adjoint_double_layer=True` for the indirect method (default) and `adjoint_double_layer=False` for the direct one.

mancellin:
Thank you @jnvn7 ! Looks good to me !

I guess we can skip the theory documentation for now until we have written that proceeding paper.
But we should still add:
- A line in `docs/changelog.rst` mentioning the new feature.
- A paragraph presenting the `method` argument in `docs/user_manual/resolution.rst`.

mancellin:
Thank you. I did a few small additions to the documentation and I think we are now good to merge.

Issue #431, opened 2023-11-11
Issue with radiation wave elevation (probably related to phases)


sdillenburg:
Dear all,

I have a question regarding the free surface elevation of radiated waves. I computed them based on the cookbook examples using the new `compute_free_surface_elevation` in version 2.0 (source: [here](https://github.com/capytaine/capytaine/blob/v2.0/capytaine/bem/solver.py#L253)) and `get_free_surface_elevation` in version 1.5 and 2.0 (source: [here](https://github.com/capytaine/capytaine/blob/v1.5/capytaine/bem/solver.py#L220) and [here](https://github.com/capytaine/capytaine/blob/v2.0/capytaine/bem/solver.py#L338))

Each function yields different wave elevations (code to generate the plots is at the end of this issue):

1.
![free_surface_dir_90__version_1 5](https://github.com/capytaine/capytaine/assets/45045610/793d5285-934c-4191-b97d-aa2afd0e80d9)
2.
![free_surface_dir_90__version_2 0](https://github.com/capytaine/capytaine/assets/45045610/c92738e1-85ae-4bb9-8513-208949ae7e89)
3.
![free_surface_version_dir_90_2 0_new_functions](https://github.com/capytaine/capytaine/assets/45045610/42cefb29-9048-4703-b4a9-9b42f35dfc27)
4.
![free_surface_dir_270__version_1 5](https://github.com/capytaine/capytaine/assets/45045610/65ef5d1a-582e-43f4-88b1-d7f410206471)
5.
![free_surface_dir_270__version_2 0](https://github.com/capytaine/capytaine/assets/45045610/7ee24aae-585c-45ab-91b2-36761998c249)
6.
![free_surface_version_dir_270_2 0_new_functions](https://github.com/capytaine/capytaine/assets/45045610/8b635a01-a4a2-4185-8be9-203ae3f7de57)

[...]

PR #434, opened 2023-11-13
Corrected boat_animation example

sdillenburg:
Fixed issue with radiated wave elevation (see #431), removed unused import, and removed non-ascii character (pi)

### DIFF
--- a/docs/user_manual/examples/boat_animation.py
+++ b/docs/user_manual/examples/boat_animation.py
@@ -1,4 +1,4 @@
-from numpy import pi as pi
+from numpy import pi
 
 import capytaine as cpt
 from capytaine.bem.airy_waves import airy_waves_free_surface_elevation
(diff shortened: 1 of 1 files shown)


docs/user_manual/examples/boat_animation.py
```diff
@@ -1,5 +1,4 @@
-import numpy as np
-from numpy import pi as π
+from numpy import pi as pi
```
mancellin, line 1: I liked my fancy π! But okay, I understand that it may be causing issues in some setup.
I guess `as pi` is not necessary if we don't rename it.

sdillenburg: > I guess `as pi` is not necessary if we don't rename it.

Ehm, yes, of course... I replaced all `π` with `pi`

mancellin:
Thank you!

Issue #266, opened 2022-12-12
Do not cache both K and its LU decomposition


mancellin:
Since #235, the default setting of the code is to cache the matrices $S$ and $K$ and the LU decomposition of $K$. It means a 50% increase in the RAM usage, which can be an issue for larger problems.

It would make more sense to keep only the LU decomposition of $K$.

PR #436, opened 2023-11-16
Preconditioner 

marcogambarini:
I have implemented a block-Jacobi, coarse-correction preconditioner on top of the hierarchical matrix engine. It can be useful for simulating large arrays of closely packed objects.


mancellin:
Thank you for this, Marco!

My main issue is with the `PrecondBEMSolver` class. I would prefer that matrix stuffs (like the preconditioning) only affects the matrix engine and does not spill out on the rest of the solver.

If I'm reading this correctly, the modifications of the `solve` function are
- Passing `leaf_order` and `path_to_leaf` from the body to `build_matrices`,
- Passing preconditioning information from `build_matrices` to the `linear_solver`.

For the former, could you attach the clustering information to the mesh instead? Other clustering information are currently stored in the mesh data structure (`Mesh` as leaves and `CollectionOfMeshes` as nodes). Do you need it to be linked to the `FloatingBody`?

For the latter, the `K` variable returned by `build_matrices` is only consumed by the `linear_solver`. So since you control both, you could make it anything, even if it is not only the matrix. It could be for instance, a tuple `K = (matrix, preconddata)`. Or better, since the matrix is likely a `BlockMatrix` object anyway, you could just attach the preconditioning data to the class instance. (It gives me some ideas for #266)

marcogambarini:
Thank you for the suggestions! I have tried to implement them in the last commit. It is much cleaner like this.

mancellin:
Thank you Marco. There is another modification I'd like to suggest.

`cluster_bodies` seems to be serving two purposes:
- Create a tree structure with the bodies using Scipy's hierarchy module (thanks for making me aware of this useful module).
- Compute and save some metadata about a tree structure.

Do you think the latter could be extracted to be part of `CollectionOfMeshes` as it may be useful for other purposes? Could it be computed in a recursive way, akin of the [`tree_wiew` method](https://github.com/marcogambarini/capytaine/blob/df6e9333c177dca7dee27c94797b8e33c0bc8eb8/capytaine/meshes/collections.py#L90-L100)?

marcogambarini:
I think tree_view does something analogous to the `dendrogram` command of scipy, in that both plot the tree structure. But I am not sure how the other metadata (`leaf_order` and `path_to_leaf`) could be built/handled recursively.

mancellin:
I took a deeper look into the code. Here are what I understand:

`leaf_order` has two purposes:
- let the user know how the `bodies` passed to `cluster_bodies` have been reordered by the clustering
- be used with `path_to_leaf` which is stored in the order of the original `bodies` list and not in the order of the actual cluster.

As far as the preconditioner is concerned, the original order of the `bodies` list is irrelevant and the `path_to_leaf` could just be ordered in the order of the actual cluster.
Such a `path_to_leaf` in the actual cluster order could be computed recursively as drafted below:

```python
class CollectionOfMeshes:
    # [...]

    def path_to_leaf(self):
        ptl = []
        for (i_submesh, submesh) in enumerate(self.submeshes):
            for p in submesh.path_to_leaf():
                ptl.append([i_submesh] + p)
        return ptl
```

In this way, we could slightly simplify the code of the preconditioner by removing references to `leaf_order` and make the preconditioner compatible with cluster trees computed by another way than `cluster_bodies`.

capytaine/bem/engines.py
```diff
@@ -315,3 +327,144 @@ def get_col_func(j):
...
+                Nrows = Aloc.all_blocks[jj, jj].shape[0]
+
+                if jj==0:
+                    v = v[:Nrows]
+                    #try-except block needed because of different function
+                    #definitions for rmatmul in np.ndarrays, blockmatrix and lowrank
+                    try:
+                        w = Aloc.all_blocks[0,1].__rmatmul__(v)
+                    except:
+                        w = Aloc.all_blocks[0,1].rmatvec(v)
```
mancellin, line 432: If `__rmatmul__` is missing somewhere (`BlockMatrix` in `block.py`?), maybe we should just implement it there to avoid the complexity here.

capytaine/bem/engines.py
```diff
@@ -315,3 +327,144 @@ def get_col_func(j):
...
+        self.linear_solver = linear_solvers.solve_precond_gmres
+
+        self.exportable_settings = {
+            'engine': 'HierarchicalToeplitzMatrixEngine',
+            'ACA_distance': ACA_distance,
+            'ACA_tol': ACA_tol,
+            'matrix_cache_size': matrix_cache_size,
+        }
+
+    def access_block_by_path(self, matrix, path):
```
mancellin, line 362: Since this does not depends on the `MatrixEngine` (`self` is unused), this could just be a method of `BlockMatrix`.

capytaine/bem/engines.py
```diff
@@ -315,3 +327,144 @@ def get_col_func(j):
...
+    ----------
+    ACA_distance: float, optional
+        Above this distance, the ACA is used to approximate the matrix with a low-rank block.
+    ACA_tol: float, optional
+        The tolerance of the ACA when building a low-rank matrix.
+    matrix_cache_size: int, optional
+        number of matrices to keep in cache
+    """
+
+    def __init__(self, *, ACA_distance=8.0, ACA_tol=1e-2, matrix_cache_size=1):
```
mancellin, line 345: Suggestion: replace with something like
```python
super().__init__(self, ACA_distance=ACA_distance, ACA_tol=ACA_tol, matrix_cache_size=matrix_cache_size)
self.linear_solver = linear_solvers.solve_precond_gmres
```

marcogambarini:
Ok, sorry, now I understood what you meant! I tried to apply these fixes in the last commit.

mancellin:
I've played a bit with the example. The caveats of Capytaine's H-matrices apply here (the dense solver is usually faster) but I think it is worthwhile to have this preconditioner as an example of advanced applications of Capytaine.

Since the scripts in the `example` folder are run in CI for every new commit, I'd like to make the example problem a bit smaller, but then the effect of the preconditioner becomes less visible. I'm still not sure how to handle this.

The PR is also missing a bit of documentation.
- Could you add a line to `docs/changelog.rst` describing this patch?
- Could you also rename `bJac_cc` with a more descriptive name and add a docstring explaining its inputs?
- It would be nice also to have unit tests for `path_to_leaf`, `access_block_by_path` and `cluster_bodies`.

marcogambarini:
> I'd like to make the example problem a bit smaller, but then the effect of the preconditioner becomes less visible.

Part of the reason why the effect is not so visible is that the construction of the matrix and the solution of the systems are timed together, while the preconditioner shows its positive effect only on the latter (and possibly a small negative effect on the former due to the construction of the coarse space).

> The PR is also missing a bit of documentation.

Do you mean on the user and theory manual?

> It would be nice also to have unit tests for path_to_leaf, access_block_by_path and cluster_bodies.

Ok, I will think about these. For now, I have implemented the other fixes.

mancellin:
Thank you, the latest commits look good to me.

>> The PR is also missing a bit of documentation.
> Do you mean on the user and theory manual?

Actually, don't worry too much about that. The whole hierarchical matrix code of mine in Capytaine is greatly under-documented and I can't ask you to document your PR without this documented as a basis.

mancellin:
Thank you Marco. I've added some basic tests, so I think we are good to merge.

PR #439, opened 2023-12-05
Expand interpolation table and add tests

jnvn7:
@mancellin I have implemented the functions/parameters to control the tabulation density in this PR. The table range and numbers of points can be defined with 'tabulation_nr', 'tabulation_nz', 'rmax', and 'zmin'. I have added some tests (pytest/test_bem_green_functions_table_density.py) for your reference on the usage. Please review and we can discuss more during the meeting. 

Thanks,
JN


mancellin:
Merging with master (with direct solver) broke https://github.com/capytaine/capytaine/blob/4cd198d572cc221de585ffcbecdfb31dfca9da34/pytest/test_bem_linear_combination_of_dofs.py#L12
for some reason.

pytest/test_bem_linear_combination_of_dofs.py
```diff
@@ -5,11 +5,11 @@
 import capytaine as cpt
 import pytest 
 
-solver = cpt.BEMSolver()
 method = ['indirect','direct']
 
 @pytest.mark.parametrize("method", method)
 def test_sum_of_dofs(method):
+    solver = cpt.BEMSolver()
```
mancellin, line 12: That's weird. Reinitializing the solver at the beginning of the test should not change its behavior.

mancellin: I think I understand the issue. It is unrelated with this PR, it comes from #420, I'll open another PR for this.

mancellin: See #442

mancellin:
I've been testing another way to switch between the legacy tabulation and the new one.
Also, I tried to rename some variables in the fortran code.

Do I understand correctly that you tuned the r_spacing such that 81/676 points are in the exponential section below $r=1$ and the rest in the linear section above $r=1$?

jnvn7:
@mancellin yes, that is correct. I was following Nemoh's approach. 81 pts are in the lower exponential region and the rest are linear.

mancellin:
I have a small issue with this exponential part: the resolution near 1 is not increasing when we increase the number of points.
```python
for nr in [1000, 1500, 2000]:
    r_range = gf.fortran_core.delhommeau_integrals.default_r_spacing(nr, 100.0, 1)
    print(r_range[(r_range > 0.1) & (r_range < 1.0)])  # Print points between 0.1 and 1

# [0.1        0.12589252 0.15848924 0.19952632 0.2511887  0.31622776
#  0.39810708 0.50118703 0.6309576  0.79432839]
# [0.1        0.12589265 0.15848924 0.19952632 0.25118843 0.31622776
#  0.39810753 0.50118703 0.6309576  0.79432756]
# [0.1        0.12589265 0.15848924 0.19952632 0.25118843 0.31622776
#  0.39810753 0.50118703 0.6309576  0.79432756]
```

It might be cleaner to do something like `np.logspace(-10.0, 0.0, int(81/676*nr))`

jnvn7:
Yea, that might actually be a better option.

mancellin:
Hi Jessica,

I'll merge the PR to move forward. Feel free to open another one if you have the time to do the small documentation update that we mentioned during a previous discussion.

PR #442, opened 2023-12-13
Fix issue with caching of matrices since #420

mancellin:
An issue noticed in #439 was actually a missed edge case from #420. The reason why it was not detected by the test suite in #420 is a mystery to me.

The cause of the issue is that the following lines
https://github.com/capytaine/capytaine/blob/4cd198d572cc221de585ffcbecdfb31dfca9da34/capytaine/bem/solver.py#L99-L113
call the function `build_matrices` which is memoized using a custom LRU cache.
But the custom LRU cache was ignoring keyword arguments, hence `build_matrices` could return the same cached value both when `adjoint_double_layer=True` and `adjoint_double_layer=False` when the same solver is used to solve the same problem with both direct and indirect methods.

This PR fixes this issue by adding support for keyword arguments in the custom LRU cache.
It also adds some test for the custom LRU cache.

### DIFF
--- a/capytaine/bem/engines.py
+++ b/capytaine/bem/engines.py
@@ -14,7 +14,7 @@ from capytaine.matrices import linear_solvers
 from capytaine.matrices.block import BlockMatrix
 from capytaine.matrices.low_rank import LowRankMatrix, NoConvergenceOfACA
 from capytaine.matrices.block_toeplitz import BlockSymmetricToeplitzMatrix, BlockToeplitzMatrix, BlockCirculantMatrix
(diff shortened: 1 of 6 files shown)


mancellin:
That test
https://github.com/capytaine/capytaine/blob/4cd198d572cc221de585ffcbecdfb31dfca9da34/pytest/test_bem_solver.py#L138-L143
was passing because of the bug in the cache, apparently.

mancellin:
Could be a good opportunity to rename the function `delete_first_lru_cache` into something clearer like `lru_cache_with_strict_maxsize`.

Issue #465, opened 2024-02-21
f2py misses some Fortran modules in Numpy 1.26.3


mancellin:
When using Python 3.12, the command
```
pip install --no-build-isolation .
```
does not create a valid package, since some Fortran module are not included anymore:
```
AttributeError: module 'capytaine.green_functions.libs.Delhommeau_float64' has no attribute 'delhommeau_integrals'
```

This is the root cause of https://github.com/conda-forge/capytaine-feedstock/pull/36 blocking #445.

Note that 
```
pip install .
```
still works as expected.

Numpy 1.26
meson 1.3.2
meson-python 0.15.0
pip 24.0

mancellin:
The compiled package has only some modules missing.

```
    This module 'Delhommeau_float64' is auto-generated with f2py (version:1.26.4).
    Functions:
    Fortran 90/95 modules:
      floating_point_precision --- pre  constants --- zero,one,pi,ii  old_prony_decomposition --- ff(),lisc(),expors(),mcas(),housrs(),sprbm(),spqfb()  matrices --- is_infinity(),build_matrices().
```

mancellin:
It seems that my 3.11 and 3.12 environments where not exactly identical. The issue might come from Numpy 1.26.2 vs 1.26.4

mancellin:
Cf. https://github.com/numpy/numpy/issues/25867

PR #467, opened 2024-02-23
Temporary upper bound on build-time Numpy version

mancellin:
To work around #465.

In pyproject.toml, the build-time Numpy version is "oldest-supported-numpy" which follow this contraint on all current Python versions.


mancellin:
Also doing another update to build version: recent versions of Numpy (1.2*) ABI are backward and forward compatible, so for recent Python version, "oldest-supported-numpy" is not required anymore, everything can be built with a recent version of Numpy (but not the latest 1.26.4 because of #465).

Whenever Numpy 2.0 is released, it should be used to build packages for Numpy 1.2* and 2.*, because packages built with 1.2* are not forward compatible with Numpy 2.* runtime.
Until #465 is fixed, Capytaine cannot be built on Numpy 2 so cannot be run on Numpy 2, so I'm adding an upper bound on Numpy runtime.

PR #472, opened 2024-03-08
Fix bug with manually defined center as tuple.

mancellin:
Namely, when the rotation_center or the center_of_mass are manually defined as something else than a vector. It was a poor design decision to support in early version of Capytaine, but now I'll try to keep it working, even if I don't recommend it.


mancellin:
Cf. https://github.com/capytaine/capytaine/issues/433#issuecomment-1985630576

PR #475, opened 2024-03-12
Remove compile-time switch between Delhommeau and XieDelhommeau

mancellin:
Compile only one version and use a run-time boolean to switch between the two.
The compile-time version was hard to debug and maintain. Hopefully this version would be easier to work with and improve.


mancellin:
Todo: need a better variable name than "legacy_delhommeau" (especially since there is already a "legacy" tabulation grid).

PR #454, opened 2024-01-25
Save tabulation on disk at initialization.

mancellin:
Replaces #404.
Finally, I believe saving on disk at the first run is more pragmatic.

PR #476, opened 2024-03-18
Tabulation parameters: default values and display

mancellin:
- Add missing info in documentation
- Better `__str__` and `__repr__`
- Change default tabulation size to 676x373 with 1000 points for the theta-integrals.


mancellin:
Should have waited for #454 before testing that.

PR #477, opened 2024-03-22
Lid generation for irregular frequencies removal

vanditya:
First step, integrating function to create internal lid


mancellin:
Thank you for the PR!
Most of my comments are cosmetic details.
We can discuss the design of the user interface in #479

mancellin:
Sorry for making your life with git more complicated, but I would suggest we focus on the lid generation in this PR and work on the usage in the solver in another PR (subsequently, or in parallel), to have smaller easier-to-handle changes in the final code.

vanditya:
I have problem adapting the generate lid function from mesh because of the need of mesh when it's already cut with the free surface. I suggest we still put it on the body which woks in this commit. However, body creation is repetitive... I need your help with it 

```
cylinder = cpt.FloatingBody(cylinderMesh)
cylinder.keep_immersed_part(free_surface=0.0,water_depth=np.infty)
lid = cylinder.generate_lid(z=-5,info=False)

body = cpt.FloatingBody(mesh=cylinder.mesh, lid_mesh=lid, dofs=cpt.rigid_body_dofs())
body.show()
```

vanditya:
generate lid implemented in the mesh. Forget my last comment I figured it out to do the keep immersed later

vanditya:
Example of irregular frequency removal as in 
On the Irregular Frequencies Appearing in Wave Diffraction-Radiation Solutions 
Malenica and Chen 1998 - OMAE

file: examples/irregular_frequency_removal.py 

results: 
lid_capytaine

mancellin:
My bad, I had written comments the other day, but had forgotten to click on "Submit", so only me could see them.
Many of them are outdated now, but some are still relevant.

mancellin:
Beyond the superficial comments above, I have two issues with the current design:

- When generating a lid below the free surface, shouldn't the lid stay inside the body? See example below (exacerbated on purpose):

```python
mesh = cpt.AxialSymmetricMesh.from_profile(lambda z: (z + 1.0)**2, np.linspace(-1.0, 0.0, 10)).merged()
lid_mesh = mesh.generate_lid(z=-0.5)
```

Figure_1

---

- In the latest version, the mesh and the lid mesh are merged early when initializing the body.
https://github.com/capytaine/capytaine/blob/58cb8baa24a2fa0212b720aa7c4fdd329bca615b/capytaine/bodies/bodies.py#L72-L76
As a consequence, all the following operations on the mesh are also done on the mesh of the lid.
It includes defining the dofs (but the lid should not move) and integrating the force on the body (but the pressure on the lid should not be taken into account).
The latter is clearly noticeable when computing the Froude-Krylov force:
```python
from capytaine.bem.airy_waves import froude_krylov_force
mesh = cpt.mesh_sphere().immersed_part()
body_without_lid = cpt.FloatingBody(
            mesh=mesh,
            dofs=cpt.rigid_body_dofs()
            )
pb_without_lid = cpt.DiffractionProblem(body=body_without_lid)

lid_mesh = mesh.generate_lid(z=-0.1, info=False)
body_with_lid = cpt.FloatingBody(
            mesh=mesh,
            lid_mesh=lid_mesh,
            dofs=cpt.rigid_body_dofs()
            )
pb_with_lid = cpt.DiffractionProblem(body=body_with_lid)

fk_with_lid = froude_krylov_force(pb_with_lid)["Heave"]
# (42923.325773727556-3.694822225952521e-13j)
fk_without_lid = froude_krylov_force(pb_without_lid)["Heave"]
# (26957.61044096423-1.8474111129762605e-13j)
```
It does not show up in your example above, because the pressure on the lid is often near zero by design.

I believe it could be avoided by merging `mesh` and `lid_mesh` only in `BEMSolver.solve` when computing the influence matrices.

mancellin:
@vanditya I pushed a commit on the `irregularFrequencyRemoval` branch of your fork that merges the support of `lid_mesh` that I implemented in other PRs such as #521.

mancellin:
I did a few cosmetic updates and fixed the issue I mentioned above:
lid_inside_mesh

Will do a few more tests before merging.

mancellin:
The lid generation is failing for non-simply-connected meshes, such as a torus or a multibody case.

Capture d’écran_2024-06-06_13-21-46

Not sure if we should fix this now or just add a warning in the documentation and investigate it later.

mancellin:
Working on #554 to, at least provide a warning to users if `generate_lid` is not expected to work.

mancellin:
I still have some interrogations about the plane geometry methods.
For instance, the following code 
```python
def on_line(l1, p):
    # Check whether p is on the line or not
    if (
        p.x <= max(l1.p1.x, l1.p2.x)
        and p.x >= min(l1.p1.x, l1.p2.x)
        and (p.y <= max(l1.p1.y, l1.p2.y) and p.y >= min(l1.p1.y, l1.p2.y))
    ):
        return True
    return False
```
is not the test whether a point is on a line.
I'm not sure if this is an implementation or a documentation issue, but I'll need to look a bit deeper into this.

mancellin:
I'm drafting another approach to this problem. Instead of looking for the inside of the waterline 2D polygon, I count the number of panels below a given point. Basically, it would still be ray-tracing but in the vertical direction.

**Pro**: no need to create a new polygon and order its vertices in counterclockwise order, thus works out-of-the-box for all watertight meshes including non-connected (multibody) cases.

**Con**: Higher computational cost as each candidate lid panels needs to be checked with all the panels in the hull mesh (but could be optimized in the future).

mancellin:
Hi @vanditya 
With all my edits, I made your work barely recognizable. Sorry about that.
I'm satisfied with the final result and would like to merge it with us both as co-authors.
Let me know if you have any issue with that.

vanditya:
Hi @mancellin, sorry for not contributing much on this due to unexpected works here. With respect to the code that underwent some changes, it is totally fine with me. Looking forward to using it! Thank you also for citing me  as co-authors of this development! Anyway I talked to some people in OMAE who are happy with this update :)

mancellin:
Thank you @vanditya.
By the way, do you have the full reference of your OMAE paper for me to refer to in the documentation ?

vanditya:
Hi @mancellin, 

It is not published online yet. Looking from last year edition, it is going to be available somewhere in September. Nevertheless, could we have a tentative citation like below?

Herdayanditya, I., Baert, B., Stempinski, F., Rauwoens, P., & Lataire, E. (2024, June). Application of Irregular Frequency Removal Feature in Capytaine for Anode Cage Installation over a Monopile Wind Turbine In International Conference on Offshore Mechanics and Arctic Engineering (volume). American Society of Mechanical Engineers.


I will keep eye on here https://asmedigitalcollection.asme.org/OMAE/OMAE2023/volume/86830, once it's there I will let you know.

PR #478, opened 2024-03-25
Add usage for new tabulation method

jnvn7:


mancellin:
Thank you Jessica!
I did a few minor corrections and updated the default values for the default I intend to use from #476.

PR #491, opened 2024-04-08
fix: build_docs Github Action

mancellin:
Trying to fix https://github.com/capytaine/capytaine/actions/runs/8600458605/job/23566763989


mancellin:
Not sure what is happening with build_docs, but this seems to fix it.

Issue #188, opened 2022-07-29
Run hydrostatics for multiple bodies


kmruehl:
This relates to https://github.com/WEC-Sim/WEC-Sim/pull/925. I'm working on updating the WEC-Sim BEMIO examples to Capytaine v1.4, and calculating hydrostatics  directly in Capytaine. It seems that the hydrostatics cannot be calculated for multiple bodies at once, and instead must be calculated separately. 

Here's an example. I get an error if I run `bodies.hydrostatic_stiffness = bodies.compute_hydrostatic_stiffness()` directly

``` python

body1 = cpt.FloatingBody.from_file('flap.gdf') # many file types supported
body2 = cpt.FloatingBody.from_file('flap.gdf') # many file types supported

body1.center_of_mass = (0,0,-3.9)
body2.center_of_mass = (-25,0,-3.9)

body1.keep_immersed_part()
body2.keep_immersed_part()
# bodies =body1+body2
# bodies.keep_immersed_part()

body1.add_all_rigid_body_dofs()
body2.add_all_rigid_body_dofs()
# bodies.add_all_rigid_body_dofs()

body1.hydrostatic_stiffness = body1.compute_hydrostatic_stiffness()
body2.hydrostatic_stiffness = body2.compute_hydrostatic_stiffness()
# bodies.hydrostatic_stiffness = bodies.compute_hydrostatic_stiffness()

hydrostatics1 = body1.compute_hydrostatics()
hydrostatics2 = body2.compute_hydrostatics()
# hydrostatics = bodies.compute_hydrostatics()
```

Calculating the hydrostatics separately is really not an issue, except that there's no easy way to store the data. `capyData = cpt.assemble_dataset(results,hydrostatics=True)` doesn't work because `bodies` does not contain `hydrostatic_stiffness`, so the data has to be manually added (which is not ideal). My request is either to allow hydrostatics to run for multiple bodies, or provide an easier way to include each `hydrostatic_stiffness` in the output.

mancellin:
We are a bit limited by Capytaine's data model: Capytaine does not _really_ support multiple bodies.
As far as hydrodynamics is concerned, having two bodies with 6 rigid dofs each or having a single "body" with 12 generalized dofs is the same.
So `body1+body2` actually returns a single body with a larger mesh and generalized dofs.

It works well for hydrodynamics.

It works so-so for hydrostatics.

The 12-dof body can still give you a 12×12 matrix for the hydrostatic stiffness, in the following way
```python
[...]

PR #501, opened 2024-04-10
Document export of rotation center to BEMRosetta

mancellin:
This should become automatic in the future, but it would require rewriting a significant part of how the multiple bodies are stored in Capytaine (cf. #188).

FYI @izabala123
Also @dav-og, since #285

### DIFF
--- a/docs/user_manual/outputs.rst
+++ b/docs/user_manual/outputs.rst
@@ -124,6 +124,54 @@ This can be fixed by explicitly converting the strings to the right format when 
 
 See also `this Github issue <https://github.com/capytaine/capytaine/issues/2>`_.
 
+
+Saving the rotation center of rigid bodies
+-----------------------------------------
+
+Some software downstream of Capytaine, such as `BEMRosetta <https://github.com/BEMRosetta/BEMRosetta>`_, require the NetCDF file to store the rotation center of each body.
+While this is not yet done automatically by Capytaine, it can be added to the dataset manually as in the following example, which is an extension of the :doc:`quickstart` example::
+
+  import numpy as np
+  import xarray as xr
+  import capytaine as cpt
+
+  body_1 = cpt.FloatingBody(
+              mesh=cpt.mesh_sphere(center=(0, 0, 0)),
+              dofs=cpt.rigid_body_dofs(rotation_center=(0, 0, 0)),
+              center_of_mass=(0, 0, 0),
+              name="my_sphere",
+          )
+  body_1.inertia_matrix = body_1.compute_rigid_body_inertia()
+  body_1.hydrostatic_stiffness = body_1.immersed_part().compute_hydrostatic_stiffness()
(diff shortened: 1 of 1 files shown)


izabala123:
First of all, I apologise for my poor internal knowledge of Capytaine.

Having the centre of the reference frame of each body is important because it indicates with respect to which point the calculations have been made. This is mandatory for rotational degrees of freedom. By including this information in the .nc file, someone who opens the .nc file will have all the necessary information, without having to search for the source code that generated the .nc file.

In the classical Nemoh,cal file, each degree of freedom of each body has the coordinates of its centre. In the case of Capytaine, I propose different ways of saving it:
- If there is a class for each body, a vector of 3 numbers for each body.
- If not, a 3xNbodies matrix.
- If you have generalised degrees of freedom, then I would advise you to do as with Nemoh.cal, that each degree of freedom has its own centre. This way, you could store it in a 3xNdof matrix, where Ndof is the number of generalised degrees of freedom. However, in order not to make the interpretation difficult, I would ask for a simple criterion, avoiding that the generalised degrees of freedom cause confusion.

Apart from this, it would be interesting to have a centre (only x, y) of the wave reference frame, unless it is always considered to be at 0.0.

Best regards
Iñaki

mancellin:
> Having the centre of the reference frame of each body is important because it indicates with respect to which point the calculations have been made

Actually, the method that fits the most to the current data model of Capytaine would be to save the full definition of the dof as a vector on each panel (`n_dof * n_panels * 3`). It removes all ambiguity on rotational dofs, but also on flexible dofs. This is drafted in #520.

But it is a bit cumbersome to manipulate, and I'll need someday to change the internal data model to keep data about individual bodies in multiple rigid bodies setups.

> Apart from this, it would be interesting to have a centre (only x, y) of the wave reference frame, unless it is always considered to be at 0.0.

It is always $(0, 0)$. I'm not sure how useful it is to be able to change it.

izabala123:
> > Having the centre of the reference frame of each body is important because it indicates with respect to which point the calculations have been made
> 
> Actually, the method that fits the most to the current data model of Capytaine would be to save the full definition of the dof as a vector on each panel (`n_dof * n_panels * 3`). It removes all ambiguity on rotational dofs, but also on flexible dofs. This is drafted in #520.
> 
> But it is a bit cumbersome to manipulate, and I'll need someday to change the internal data model to keep data about individual bodies in multiple rigid bodies setups.

I understand. However, I was wondering if giving the freedom for each panel to have its own reference system would make any difference to the calculation. Given that the translation of forces from one system to another is so simple, I wonder if perhaps the simplest way would be for the centre of the reference system to be the same for all the bodies and panels.

> > Apart from this, it would be interesting to have a centre (only x, y) of the wave reference frame, unless it is always considered to be at 0.0.
> 
> It is always (0,0). I'm not sure how useful it is to be able to change it.

I welcome the decision. There is some solver that allows you to modify it, but it doesn't seem necessary either.

PR #533, opened 2024-05-14
feat: add internal timer in BasicMatrixEngine

mancellin:


mancellin:
Maybe it should be in the BEMSolver instead. Otherwise we miss time spent on other steps (matrix multiplication, velocity reconstruction for forward speed, etc.)

mancellin:
Abandonned for #674

PR #535, opened 2024-05-15
Updated Haskind test case according to unit amplitude convention

marcogambarini:
The Haskind test case uses the radiation potential solution, and the convention for the boundary conditions of radiation problems has been changed in 56353b6. I have corrected the test accordingly.


mancellin:
Good catch, thank you!

Issue #553, opened 2024-06-06
Allow gmsh msh v4 files to be loaded


AnthonyMcD:
Currently the load_mesh function only supports loading v2 format msh files generated by gmsh. The default msh file generated by the current version of gmsh is now v4 format, so it would be good if load_mesh could also support them.

Given that the [load_MSH](https://github.com/capytaine/capytaine/blob/master/capytaine/io/mesh_loaders.py#L791) function in capytaine is based on meshmagick, and the [meshmagick load_MSH function](https://github.com/LHEEA/meshmagick/blob/master/meshmagick/mmio.py#L914) now supports v4 format msh files it should be straightforward to insert the appropriate code into capytaine so that it can also support v4 format msh files.

I am going to attempt this enhancement locally for my own work and will submit a pull request if successful, unless there is some reason that v4 format msh files can/should not be used in capytaine.

mancellin:
Meshmagick's gmsh v4 parser is a thin wrapper around on [gmshparser](https://gmshparser.readthedocs.io). So if you're going to install a third-party library, you might as well use [meshio](https://github.com/nschloe/meshio) which support gmsh v4 and already have some bindings with Capytaine.

But I agree that supporting v2 directly and not v4 is confusing for users, so I'd be happy to see your pull request.

PR #556, opened 2024-06-07
Allow import of msh v4 file in load_mesh

AnthonyMcD:
Hi

I've added the ability to load gmsh v4 in the load_mesh function. 

As suggested in #553 I've used meshio. However as meshio is an optional dependency I have added a error message for cases where a v4 mesh load is attempted without meshio installed.  I have also maintained the code for loading v2 msh files for cases where meshio is not installed (and a v2 msh file load is attempted).

Two tests have been added, one for v2 msh and one for v4 msh


mancellin:
Merged, thank you again!

PR #596, opened 2024-09-16
build: use OpenMP in MacOS wheels

mancellin:
This is supposed to be supported by Meson now.


mancellin:
> It seems to work on MacOS-12.

MacOS-12 Github Action runner is being removed.

Issue #601, opened 2024-10-11
`connected_components` is very expensive


ldouteau:
Hi, thanks again for Capytaine!

Im trying to run a HDB (100 frequencies, 10 dofs) on a huge mesh (35k vertices, 33k faces), and the wavelength check is taking ages (7hours expected). I narrowed the issue to `solve_all` calling`_check_wavelength_and_irregular_frequencies` calling `connected_components`, which is very expensive.

https://github.com/capytaine/capytaine/blob/c5658dd430e42eb16fd7edb26f919ef28f321c14/capytaine/meshes/properties.py#L202-L233

From what i understand this function starts with a clustering step, to gather the faces that share at least 1 point. It's this step that seems to be the most expensive. E.g., from my test case:

```
clustering 22.139901876449585s
check 0.0s
face identifiction 0.11008667945861816s
face extraction 0.0314946174621582s
```

I've not had the time to investigate on this, i'll do eventually, but if someone wants to jump in feel free

mancellin:
Ok I should have tested that on bigger meshes :grimacing:

This is only to provide a warning if you might encounter irregular frequencies, so if you know what you're doing, you can turn it off, for instance with something like that:
```python
from capytaine.io.xarray import problems_from_dataset
problems = problems_from_dataset(test_matrix, body)
solver = cpt.BEMSolver()
results = solver.solve_all(problems, keep_details=False, _check_wavelength=False)
dataset = cpt.assemble_dataset(results)
```

mancellin:
The code is redoing the exact same clustering for each of the 100 freq×10 dofs problems...
That should be easy to improve.

ldouteau:
I though about this over the weekend, with 2 main conclusions

* I wonder if the generation of connected vertices could be a cached property of the mesh instance ? I'm pretty sure it would work for my test case, but i dont know for others
* I've though about another algorithm that introduces some vectorization. I'll test that & open a pr if the results are promizing

mancellin:
> I wonder if the generation of connected vertices could be a cached property of the mesh instance ? I'm pretty sure it would work for my test case, but i dont know for others

In #602, I'm caching the `body.first_irregular_frequency_estimate()`.
I guess we could also cache the `connected_components` of the `Mesh` (in the same way we cache the `faces_centers` or the `faces_areas` of the mesh), but at the moment I believe their only use in is `first_irregular_frequency_estimate()`.

[...]

PR #603, opened 2024-10-14
Improved algorithm for connected components

ldouteau:
Following #601, this PR optimizes the function `connected_components`. Benchmark on the test case of #601: 7x faster, without the cache optimization from #602. Should be 100% compatible with #602

Feel free to comment or modify

### DIFF
--- a/capytaine/meshes/properties.py
+++ b/capytaine/meshes/properties.py
@@ -4,7 +4,9 @@ Based on meshmagick <https://github.com/LHEEA/meshmagick> by François Rongère.
 # Copyright (C) 2017-2019 Matthieu Ancellin, based on the work of François Rongère
 # See LICENSE file at <https://github.com/mancellin/capytaine>
 
+from functools import reduce
 import numpy as np
+from numpy.typing import NDArray
 
 
 def compute_faces_properties(mesh):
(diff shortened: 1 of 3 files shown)


ldouteau:
Tests failing, most probably due to numpy version being < 2.0, see https://github.com/numpy/numpy/issues/22021#issuecomment-1918920677

mancellin:
The value error is triggered on
```python
mesh = cpt.mesh_sphere(radius=1.0, resolution=(4, 3)).immersed_part()
components = connected_components(mesh)
```

ldouteau:
Hum all tests are passing on my side (NumPy 2.1.2). I'll check with later versions

ldouteau:
> ```python
> clustering(np.array([[0, 1, 2, 3], [1, 2, 6, 7]]))
> # [array([0, 1, 2, 3])]
> # Should be [array([0, 1, 2, 3, 6, 7]) ?
> ```

Nice edge case, this one is fixed. Regarding the others i can't reproduce the failure, even with Numpy 1.26

mancellin:
I did not have the time to run a few more tests, but I will as soon as possible.

By the way, we are in the process of releasing as much as possible of Capytaine under a more permissive license than GPL (most likely Apache 2). Do you agree with your work on this PR to be release under another open-source license?

mancellin:
I did some quick benchmarks and your implementation is indeed faster, especially for large meshes with few clusters.
My implementation can compete when there is a large number (e.g. > 50) of clusters, but that is not the most common use case of Capytaine.

PR #617, opened 2024-11-22
feat: LiangWuNoblesse Green function implementation

mancellin:
Add `cpt.LiangWuNoblesseGF` that can be passed to `BEMSolver` for (infinite-depth) resolution with another implementation of the Green function.


mancellin:
TODO: documentation and benchmarks

PR #631, opened 2024-12-20
fix: re-enable optimization in matrix computation

### LINKED ISSUE #628: Disbaled optimization after #510
Some sloppy developer (me) forgot to re-enable the optimization of the wave part of the Green function after #510 

https://github.com/capytaine/capytaine/blob/686bceefdfee553b8f1c208e276ce919ea06aec6/capytaine/green_functions/libDelhommeau/src/matrices.f90#L84-L85

mancellin:
I believe it is safe.
But it would be a good opportunity to test and document it better.

Fix #628


mancellin:
The recent refactoring of the Fortran makes it difficult to re-implement this optimization so easily. Will need a bit more work.

PR #638, opened 2025-02-04
build: update dev dependencies management.

mancellin:


mancellin:
Waiting for some issues in pip and uv before moving forward with [dependency-groups].

PR #669, opened 2025-04-01
docs: add examples of a GPU solver and a benchmark of different solvers

SoundsSerious:
Introduce examples demonstrating the implementation of custom solvers, including a GPU-based solver, and provide a benchmark generator to compare solver performance across various problem sizes.

I've placed these examples in `custom_solvers` folder in `examples` to segment them from the more topical examples.

I've also included a note that GPU solvers are not officially supported by Capytaine and to get support from the 3rd party solver channel


mancellin:
Thanks for this!

I have two main comments:
- the first example seems to be based on https://github.com/capytaine/capytaine/pull/446#issuecomment-1875117715, but you did not keep the caching mechanism and the function name does not fit anymore. I believe it is important not to redo the transfer to GPU and the LU decomposition for the same matrix twice.
- the second example is not directly related to Capytaine, so I'm not sure if it make sense to include it in the Cookbook. Or it should be on actual Capytaine problems, not random matrices.

Because of another recent project, I have now access to a recent GPU and more interest in it, so I'll try to do some benchmarks myself.

mancellin:
> I believe it is important not to redo the transfer to GPU and the LU decomposition for the same matrix twice.

After testing it, the GPU actually makes the linear solver fast enough that it is almost negligible in comparison with the rest, so the optimization is not so significant.

Also, there might be a bug, the results do not match. I'm looking into that.

mancellin:
The wrong result was due to the linear solver returning a column vector (shape (n, 1)) instead of a 1D vector (shape (n,)). I'm adding a safeguard in #670.

SoundsSerious:
Congrats on the new hardware! GPU computing definitely seems to be becoming the norm with hybrid shared memory architecture becoming more normal. 

Regarding the lost LU caching mechanism since it was an example I was thinking it might be too distracting so removed it. Maybe I can add it back in as commented optimization. To your last point it didnt appear to matter much!

Regarding the benchmark maybe that's something that can be tabulated, or put in a github actions. As they say what gets measured gets improved! This project has some pretty nice benchmark results for some inspiration https://pypi.org/project/diskcache/.  This also may be useful for caching LU stores ect, its straightforward and reliable. 

I'm certainly not wetted to the format of this commit, if you'd like to change please feel free. I can also make the changes as well! 

Perhaps its best to put this in a GPU branch for now if you want to experiment? Or maybe certain parts of it, and push just the one example to master?

SoundsSerious:
> The wrong result was due to the linear solver returning a column vector (shape (n, 1)) instead of a 1D vector (shape (n,)). I'm adding a safeguard in https://github.com/capytaine/capytaine/pull/670.

Good catch! I am definitely spread too thin!

mancellin:
I reworked the example to have both with and without cache and some simple benchmarking in a single example.

SoundsSerious:
@mancellin and thanks to you for putting this project together! Really helped out with my project.

Maybe this project of ours can help you if you want to go on a validation / model improvement path. We have a pretty snazzy magnetic encoder setup that works well on smaller models down to the micro meter level. 
https://github.com/neptunyalabs/wavetankOS

PR #696, opened 2025-05-28
support freq in Hz

### LINKED ISSUE #456: Add frequency (Hz) coordinate to datasets
I often do the following after running the Capytaine solver:

```python
ds = ds.assign_coords(freq=("omega",ds['omega'].values/2/np.pi))
ds['freq'].attrs['long_name'] = 'Frequency'
ds['freq'].attrs['units'] = 'Hz'
```

which allows me to more easily make plots with the `x_label` being `Frequency [Hz]` by using `da.plot(x='freq', ...)`, e.g.,

```python

Fexc = (ds['diffraction_force'] + ds['Froude_Krylov_force']).sel(wave_direction=0)

fig, ax = plt.subplots(nrows=2,
                       sharex=True)

np.abs(Fexc).plot(x='freq', ax=ax[0], hue='influenced_dof')
ax[0].set_ylabel('$|F_e|$')

np.arctan(np.real(Fexc)/np.imag(Fexc)).plot(x='freq', ax=ax[1], hue='influenced_dof', add_legend=False)
ax[1].set_ylabel('$\\angle{ F_e } $')

for axi in ax:
    axi.set_title('')
    axi.label_outer()
    axi.grid()
    axi.autoscale(enable=True, axis='x', tight=True)
    axi.spines[['right', 'top']].set_visible(False)
    axi.set_xscale('log')
```

image

image

Looking at the source (for the first time in a while...), I think this could be handled cleanly in [`xarray.py`](https://github.com/capytaine/capytaine/blob/master/capytaine/io/xarray.py) in two ways:

1. (more complete) Add `frequency` to the `frequency_keys` -- this would allow the user to formulate problems using `frequency` in addition to the existing options.

https://github.com/capytaine/capytaine/blob/f5f0b9fb439d9d865adc1c91255b87537b9bbed1/capytaine/io/xarray.py#L77

2. (easier) Add my lines above to the [`assemble_dataset` function](https://github.com/capytaine/capytaine/blob/f5f0b9fb439d9d865adc1c91255b87537b9bbed1/capytaine/io/xarray.py#L255) to add the `frequency` coordinate to the results of any problem.

I can definitely do **2**, and probably do **1** with some guidance. Any thoughts @mancellin?

rgcoe:
Closes #456.

Note: I added a few tests where it seemed appropriate; let me know if you concur.

### DIFF
--- a/capytaine/bem/problems_and_results.py
+++ b/capytaine/bem/problems_and_results.py
@@ -59,7 +59,7 @@ class LinearPotentialFlowProblem:
                  body=None,
                  free_surface=_default_parameters['free_surface'],
                  water_depth=None, sea_bottom=None,
-                 omega=None, period=None, wavenumber=None, wavelength=None,
+                 omega=None, freq=None, period=None, wavenumber=None, wavelength=None,
                  forward_speed=_default_parameters['forward_speed'],
                  rho=_default_parameters['rho'],
                  g=_default_parameters['g'],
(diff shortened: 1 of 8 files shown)


mancellin:
That looks good, thank you!

I would have proposed naming the parameter `frequency`, but since at least *wavespectra* uses `freq`, we can keep that.

Could you skim through the documentation to add this feature where its relevant? At least the "Setting a problem" section and the "Minor update" section of the changelog.

mancellin:
Thank you!

PR #698, opened 2025-05-28
add joblib example

rgcoe:
Adds an example of using `joblib` to perform a parallel speedup study per todo below

https://github.com/capytaine/capytaine/blob/22eb486cb3c3f05325bd9ddd592b5e6800f24aa0/docs/examples/C_advanced_cookbook.rst?plain=1#L11-L14

### DIFF
--- a/docs/examples/C_advanced_cookbook.rst
+++ b/docs/examples/C_advanced_cookbook.rst
@@ -11,7 +11,10 @@ The scripts can be downloaded individually as Python files from the |examples_fo
 C1. Parallel computation with Joblib
 ------------------------------------
 
-TODO
+This example runs a parallel speedup study.
+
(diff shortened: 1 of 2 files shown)


mancellin:
Thank you for tackling this. This hits some pain points regarding the user interface of parallel computing.

The only reason scaling is not nearly linear is because all cases are already using all the threads, as you can check with e.g. your OS process manager. `Solving BEM problems with 2 threads` is misleading and the user information of what is going on should be improved.

While the *joblib* dispatch in `n_jobs` processes is usually more efficient, especially for small problems, it requires more RAM (`n_jobs` times more) and can thus cause issues with large problems (cf. https://github.com/capytaine/capytaine/issues/148 and https://github.com/capytaine/capytaine/pull/687)

PR #725, opened 2025-07-15
Export to wamit

dmesbah:
Fix: Corrected export logic to handle both omega and period dimensions consistently


mancellin:
See #727

PR #727, opened 2025-07-15
Export wamit

dmesbah:
Fix: Corrected export logic to handle both omega and period dimensions consistently

### DIFF
--- a/pytest/test_io_wamit.py
+++ b/pytest/test_io_wamit.py
@@ -3,6 +3,7 @@ import logging
 from pathlib import Path
 
 import numpy as np
+import re
 import pandas as pd
 import xarray as xr
 import pytest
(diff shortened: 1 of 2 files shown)


pytest/test_io_wamit.py
```diff
@@ -160,10 +161,123 @@ def test_export_wamit_hydrostatics(full_dataset, tmpdir):
...
+            "period": np.asarray(period_val),
+            "wave_direction": [0.0],
+            "radiating_dof": list(immersed_body.dofs),
+            "water_depth": [np.inf],
+            "rho": [1025],
+        }
+    )
+
+    solver = cpt.BEMSolver()
+    ds_period = solver.fill_dataset(test_matrix_period, immersed_body)
```
mancellin, line 267: It should be sufficient to do something like
```
ds_period = ds_omega.swap_dims({'omega': 'period'})
```

PR #782, opened 2025-11-17
Fix timer parallel

julia-schmitt:


mancellin:
Thanks, looks good!

I'm wondering if we could do even better and actually update the main `solver.timer` such that we can still access it later, for instance:
```python
cpt.set_logging("WARNING")  # Disable all infos, including default printing of timer summary
[...]
results = solver.solve_all(pbs, n_jobs=2)
print(solver.timer_summary())  # Print manually timer summary
```
but that might mean redesigning the Timer object https://github.com/capytaine/capytaine/blob/28fe933d3e117966986fa2a893602c794ea1b0c8/src/capytaine/tools/timer.py to account for the different processes.

mancellin:
or maybe initialize `solver.timer` with lists of `Timer`s
```
{"Solve total": [Timer()], "  Green function": [Timer()], "  Linear solver": [Timer()]}
```
and add more `Timer`s to the list after a parallel resolution.

mancellin:
or maybe a [`defaultdict`](https://docs.python.org/3/library/collections.html#collections.defaultdict):
```python
self.timer = {"Solve total": defaultdict(Timer), "  Green function": defaultdict(Timer), "  Linear solver": defaultdict(Timer)}
...
# In solve:
with self.timer["  Green function"][0]:
    ...
# In solve_all parallel:
for grp_results, timer in groups_of_results:
    results.extend(grp_results)
    for timer_key in self.timer:
        self.timer[timer_key][id_process_that_solved_that_group].add_data_from_other_timer(timer[timer_key][0])
```
where we'd also need to implement `add_data_from_other_timer` in the `Timer` class.

PR #784, opened 2025-11-18
Ram estimation

julia-schmitt:


mancellin:
Seems to be working great, thank you!

I would suggest to uniformize the two messages:
```
Estimated peak RAM usage: 2 GB
Actual peak RAM usage: 1.58 GB
```
without a decimal point in the estimatation and without displaying the `ram_limit`.

mancellin:
```
[12:11:44] INFO     Estimated peak RAM usage: <1 GB.                                                                                                        
           INFO     To get the RAM consumption the dependency 'psutil' is necessary.                                                                        
Solving BEM problems with 2 threads: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:03
[12:11:48] INFO     Actual peak RAM usage: None GB.                                                                                                         
```
would be nicer as
```
[12:11:44] INFO     Estimated peak RAM usage: <1 GB.                                                                                                               
Solving BEM problems with 2 threads: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:03
[12:11:48] INFO     Actual peak RAM usage: Not measured since optional dependency `psutil` cannot be found.                                     
```

mancellin:
Thanks. I'll update the documentation about the new optional dependency.

PR #787, opened 2025-11-24
feat: clipping for new mesh class

mancellin:


mancellin:
Missing mechanism to pass the ids of clipped faces to the body to clip the dofs if necessary. Could be fixed simply by associating some kind of metadata to the faces of the mesh instead.

PR #799, opened 2025-12-15
feat: visualization of bodies (including dofs)

mancellin:


mancellin:
Remaining tasks: #874

PR #803, opened 2025-12-19
dispatch threads and processes

julia-schmitt:


src/capytaine/bem/solver.py
```diff
@@ -373,8 +373,30 @@ def _check_wavelength_and_irregular_frequencies(problems):
...
+        if n_cpu is None:
+            psutil = silently_import_optional_dependency("psutil")
+            if psutil is not None:
+                n_cpu = psutil.cpu_count(logical=False)
+
+        nb_threads_per_jobs = int(n_cpu // max_jobs) if n_cpu is not None else None
+        if nb_threads_per_jobs == 0:
+            nb_threads_per_jobs = None
+
+        return max_jobs, nb_threads_per_jobs
```
mancellin: We could return a `dict`, such that we could do
```python
results = solver.solve_all(problems, **solver.dispatch(problems, memory=10, n_cpu=4))
```

Later, we will maybe remove the duplicate reference to `solver` and `problems`.

PR #808, opened 2026-01-05
fix bug in lu overwrite

julia-schmitt:


pytest/test_bem_engines_basic_matric_engine.py
```diff
@@ -107,6 +107,15 @@ def test_all_linear_solvers_work():
     )
 
 
+def test_lu_overwrite():
+    K = np.array(np.random.rand(100,100), order='F')
+    K_copy = np.copy(K)
+    engine = cpt.BasicMatrixEngine(linear_solver="lu_decomposition_with_overwrite")
+    engine.last_computed_matrices = (None, K)
+    engine.linear_solver(K, np.random.rand(100))
+    assert np.all(K_copy != K)
```
mancellin: `np.any` might be sufficient.

PR #816, opened 2026-01-12
Issue 783 (Failed resolution error message is not displayed when solving in parallel) 

julia-schmitt:


mancellin:
Would be nice if it could be displayed immediately after the group of problem is skipped. Not sure how much more complicated it would be.

PR #820, opened 2026-01-19
fix: correct frequency type for kochin_data_array

julia-schmitt:


mancellin:
Thank you. Could you please add an entry in `docs/changelog.rst` in the "Bug fixes" section and also add at least one test in the test suite?

PR #871, opened 2026-04-08
Near field (single rigid body) mean drift force

julia-schmitt:


src/capytaine/post_pro/mean_drift_force.py
```diff
@@ -71,4 +83,186 @@ def far_field_mean_drift_force(X, dataset):
...
+    mesh = body.mesh
+    rho = results[0].rho
+    g = results[0].g
+    omega = rao.coords["omega"].values
+    nb_dir = rao.sizes["wave_direction"]
+    zero_block = np.zeros((nb_dir, 3, 3))
+    zero_33 = zero_block[0, ...]
+
+    translation = rao.sel(radiating_dof=["Surge", "Sway", "Heave"]).values[0, :, :] # translation[i_dir, xyz]
+    rotation = rao.sel(radiating_dof=["Roll", "Pitch", "Yaw"]).values[0, :, :] # rotation[i_dir, xyz]
```
mancellin: For generalization to multibody, I'd like to be able to consider all dofs as [`AbstractDof`](https://github.com/capytaine/capytaine/blob/3c818103a9a5870827a4f964aae8eb9e99380ff2/src/capytaine/bodies/dofs.py#L16-L46).
You can recover `motion_order1` with the sum of `evaluate_motion_at_points` method on each dof.

But `rotation_matrix` and `translation_matrix` are trickier. We might need to implement other methods in the `AbstractDof` base class. Maybe we can start by not relying on their name, but whether the dof is a subclass of `TranslationDof` or `RotationDof`.

src/capytaine/post_pro/mean_drift_force.py
```diff
@@ -71,4 +83,186 @@ def far_field_mean_drift_force(X, dataset):
...
+            potential_gradient.loc[{"wave_direction": res.wave_direction}] += solver._compute_potential_gradient(mesh, res) # diffracted
+
+        elif isinstance(res, RadiationResult):
+            rao_rad = rao.sel({"radiating_dof": res.radiating_dof}).values[0, :]
+            potential_gradient += rao_rad[:, None, None] * solver._compute_potential_gradient(mesh, res)[None, :, :] # radiated
+            # potential_gradient[i_dir, i_face, xyz] = rao_rad[i_dir] * _compute_potential_gradient[i_face, xyz]
+
+    return potential_gradient
+
+def total_free_surface_elevation(solver, vertices_middle, results, rao):
```
mancellin, line 242: Maybe this could be implemented as a general method next to https://github.com/capytaine/capytaine/blob/3c818103a9a5870827a4f964aae8eb9e99380ff2/src/capytaine/bem/solver.py#L523

src/capytaine/post_pro/mean_drift_force.py
```diff
@@ -71,4 +83,186 @@ def far_field_mean_drift_force(X, dataset):
...
+            "influenced_dof": list(body.dofs.keys()),
+        }
+    )
+
+def motion_order1(points, translation, rotation):
+    results = translation[:, None, :] + np.cross(rotation[:, None, :], points[None, :, :])
+    # results[i_dir, i_face, xyz] = translation[i_dir, xyz] + cross(rotation[i_dir, xyz], point[i_face, xyz])
+    return results
+
+def total_potential_gradient(solver, mesh, results, rao):
```
mancellin, line 221: Maybe this could be implemented as a general method next to https://github.com/capytaine/capytaine/blob/3c818103a9a5870827a4f964aae8eb9e99380ff2/src/capytaine/bem/solver.py#L448

src/capytaine/post_pro/mean_drift_force.py
```diff
@@ -71,4 +83,186 @@ def far_field_mean_drift_force(X, dataset):
...
+            free_surface_elevation.loc[{"wave_direction": res.wave_direction}] += solver.compute_free_surface_elevation(vertices_middle, res) # diffracted
+
+        elif isinstance(res, RadiationResult):
+            rao_rad = rao.sel({"radiating_dof": res.radiating_dof}).values[0, :]
+            free_surface_elevation += rao_rad[:, None] * solver.compute_free_surface_elevation(vertices_middle, res)[None, :] # radiated
+            # free_surface_elevation[i_dir, i_edge_waterline] = rao_rad[i_dir] * compute_free_surface_elevation[i_edge_waterline]
+
+    return free_surface_elevation
+
+def hydrostatics_forces(body, rao, rho, g, z):
```
mancellin, line 263: Kinda duplicate of https://github.com/capytaine/capytaine/blob/3c818103a9a5870827a4f964aae8eb9e99380ff2/src/capytaine/bodies/hydrostatics.py but I'm not sure how to merge them together.

src/capytaine/post_pro/mean_drift_force.py
mancellin: The pattern appearing several times here is computing a something on a linear combination of `results`.
It might be convenient (and maybe even more performant) to define addition and multiplication of `LinearPotentialFlowResults` objects, such that the mean drift force computation would involve a single `_compute_potential_gradient(points, sum(rao[i]*result[i]))`.

src/capytaine/post_pro/mean_drift_force.py
mancellin: To be done in another future PR.

mancellin:
TODO: solve for more than a single frequency at once.

mancellin:
Fails on Python 3.8: "AttributeError: module 'numpy' has no attribute 'matvec'"

PR #887, opened 2026-04-28
feat: hydrostatics as xarray.Dataset

mancellin:

PR #898, opened 2026-05-29
refactor: export hydrostatics to Nemoh format

mancellin:


mancellin:
Waiting for #887 to complete.

PR #899, opened 2026-05-29
feat: export far-field mean drift force to WAMIT format

mancellin:
FYI @julia-schmitt

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -237,6 +237,10 @@ Bug fixes
 
 * Progress bar display progress of groups of problems solved, not single problems, leading to more accurate time estimation (:pull:`882`)
 
+* The RAO used to be computed wrongly with the transpose of the actual transfer function. (:issue:`891` and :pull:`892`)
+
+* :func:`~capytaine.io.xarray.assemble_matrices` now returns the matrices ordered along the frequency axis like the input ``results``, instead of sorted by increasing frequency. (:issue:`797`)
+
 Internals
 ~~~~~~~~~
 
--- a/docs/examples/src/A1_single_body_hydrodynamics.py
+++ b/docs/examples/src/A1_single_body_hydrodynamics.py
@@ -30,6 +30,7 @@ test_matrix = xr.Dataset(
         "omega": np.linspace(0.1, 3.0, 20),
         "wave_direction": [0.0, np.pi/2],
         "radiating_dof": list(body.dofs),
+        "theta": np.linspace(0, 2*np.pi, 100),
     }
 )
 # Water density, gravity and water depth have not been specified.
--- a/docs/examples/src/A1_single_body_hydrodynamics.py
+++ b/docs/examples/src/A1_single_body_hydrodynamics.py
@@ -30,7 +30,6 @@ test_matrix = xr.Dataset(
         "omega": np.linspace(0.1, 3.0, 20),
         "wave_direction": [0.0, np.pi/2],
         "radiating_dof": list(body.dofs),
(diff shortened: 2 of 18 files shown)


stefan-roland:
Dear @mancellin 

I’ve also been working on WAMIT mean-drift export. I already had a small helper for writing `.8/.9`-style output from a mean-drift `DataArray` that might be of interest to you.

It makes the mean drift real-valued explicitly and generalizes over DOFs through `influenced_dof`, which might be useful if support is extended beyond the current `.8` export to `.7` and `.9`

Something along these lines:

```python
def export_wamit_mean_drift(mean_drift, rho, g, filename, length_scale=1.0, wave_amplitude=1.0):
    """Write mean-drift loads to a WAMIT .8/.9 file (the dof set decides which).

    `mean_drift` is a real DataArray with dims (omega, wave_direction, influenced_dof). Follows the
    WAMIT manual section 5.2 OPTN.8/.9 column layout:
        PER  BETA1  BETA2  I  Mod(Fi)  Pha(Fi)  Re(Fi)  Im(Fi)
    Mean drift is real -> Im = 0, Mod = |F|, Pha = 0 or 180 deg. Positive indices only.
    Nondimensionalized per section 3.8:
        Fbar_i = F_i / (rho * g * A^2 * L^k),  k = 1 for forces (1,2,3), k = 2 for moments (4,5,6).
    """
    dofs = list(mean_drift.coords["influenced_dof"].values)
    betas = np.atleast_1d(mean_drift["wave_direction"].values)
    omegas = np.atleast_1d(mean_drift["omega"].values)
    periods = 2 * np.pi / omegas

    order = np.argsort(periods)

    with open(filename, "w") as f:
        for idx in order:
            period = float(periods[idx])
            if not np.isfinite(period) or np.isclose(period, 0.0):
                continue
            omega = float(omegas[idx])
            for beta in betas:
                beta_deg = np.degrees(float(beta))
                for dof in dofs:
                    i_dof = DOF_INDEX[dof]
                    k = 1 if DOF_TYPE[dof] == "trans" else 2
                    force = float(np.real(
                        mean_drift.sel(wave_direction=beta, omega=omega, influenced_dof=dof).item()
                    ))
                    f_nd = force / (rho * g * wave_amplitude**2 * length_scale**k)
                    mod = abs(f_nd)
                    pha = 0.0 if f_nd >= 0.0 else 180.0
                    f.write(
                        "{:12.6e}\t{:12.6f}\t{:12.6f}\t{:5d}\t{:12.6e}\t{:12.3f}\t{:12.6e}\t{:12.6e}\n".format(
                            period, beta_deg, beta_deg, i_dof, mod, pha, f_nd, 0.0
                        )
                    )
    print(f"Saved WAMIT mean-drift file: {filename}")
```
I’m not sure whether that fits the current API/design direction here, but if useful, I can open a small follow-up PR/refactor based on this idea.

mancellin:
Thank you for sharing this.

I guess the main difference is that you stored the near-field and far-field in the same xarray structure, which make it easier to reuse the same export method.
Right now we store the far-field one as three variables instead of a single one depending on "influenced_dofs", because I was worried of xarray issues when having in the same dataset some arrays with different "influenced_dofs" sizes.
But it would make sense to have something like
```python
def export_wamit_8(dataset, filename, ...):
    return export_wamit_mean_drift(merge_far_field_mean_drift_variables(dataset), filename, ...)

def export_wamit_9(dataset, filename, ...):
    return export_wamit_mean_drift(dataset["near_field_mean_drift_force"], filename, ...)
```

PR #901, opened 2026-06-03
far field bi-directional mean drift force

julia-schmitt:


mancellin:
TODO: add some tests, at least checking that the code executes and returns a dataarray of the right shape.

PR #902, opened 2026-06-16
Symmetry for S and K matrices

julia-schmitt:


src/capytaine/bem/engines.py
```diff
@@ -137,7 +137,7 @@ def build_S_matrix(self, points, mesh, **gf_params) -> np.ndarray:
             List of collocation points. Usually went through `_normalize_points`.
         """
         def build_S_rows(slice_):
-            S, _ = self.green_function.evaluate(points[slice_, :], mesh, **gf_params)
+            S, _ = self.green_function.evaluate(points[slice_, :], mesh.merged(), **gf_params)
```
mancellin: Do the merge once outside of `build_S_rows`. To avoid redoing the same computation for each row of the matrix.

PR #909, opened 2026-07-02
feat: WAMIT export near field mean drift force

mancellin:
Waiting on #871


mancellin:
To be rebased on #871 once it is done.

PR #922, opened 2026-07-16
fix: do not override existing logging

mancellin:


jynahon:
Thanks @mancellin :+1:

